PR #20, opened 2024-09-27
Prove equations 41 and 46 are equivalent

Shiney:
I did this as two implications as no other example of equivalence yet


pitmonticone:
Thanks!

PR #23, opened 2024-09-27
Solutions for 53,905 cases by substituting+renaming variables

carlini:
This branch introduces 19,000 solutions that can be found via simple renaming of variables with substitution.

This should not be merged until it's decided how the final structure of the project will coordinate the layout of proofs. As such I haven't added any other files to the workflow while that is all decided here.

Just committing these for now to avoid duplication of work for these cases.


carlini:
I've updated the algorithm and it can now produce 53,905 total proofs correctly. Still waiting on a decision for the restructure before cleaning up.

Shreyas4991:
Could you add the `do not merge` Label to this PR?

Command-Master:
Before merging it I think it should be transitively reduced

carlini:
@Command-Master Ah yes you're obviously right. This reduces the number of files considerably.

carlini:
@Shreyas4991 This PR is now ready to be looked at.

teorth:
Thanks for the contribution!  I've added a mention of it to the front page README.md and also added a blueprint chapter to https://github.com/teorth/equational_theories/blob/main/blueprint/src/chapter/simple_rewrites.tex .  When you have a chance, can you update the blueprint to describe how the implications were generated?  Thanks!

teorth:
Also, can you place your simple_rewrites folder inside the Generated subfolder?

Issue #33, opened 2024-09-28
SUBGRAPH: Prove that 39 and 42 are equivalent


teorth:

Command-Master:
This isn't true---the naturals with `x ∘ y = y` satisfy 39 (`x ∘ x = y ∘ x`) but not 42 (`x ∘ y = x ∘ z`)

teorth:
Oh, good point, that was silly on my part.

PR #42, opened 2024-09-28
Two more implications from equation 5

gl3nnleblanc:
Two more simple implications from equation 5 that became obvious to me after reading [#33](https://github.com/teorth/equational_theories/issues/33#issuecomment-2380356699)


gl3nnleblanc:
I suspect this is covered by #23, but will leave this open anyway :)

PR #44, opened 2024-09-28
Auto-generated proofs for all laws trivially implying Eq2 (singleton) and Eq46 (constant)

amirlb:
This is done with a two large files, not with any kind of introspection on equation structure.

815 theorems are included for singletons and 85 for constant laws.

### DIFF
--- a/equational_theories/Generated/Constant.lean
+++ b/equational_theories/Generated/Constant.lean
@@ -1,0 +1,267 @@
+import equational_theories.Magma
+import equational_theories.AllEquations
+import Mathlib.Tactic
+
+namespace Constant
+
+/- Equational laws that imply the operation is a constant function -/
+
+theorem Equation41_implies_Equation46 (G: Type*) [Magma G] (h: Equation41 G) : Equation46 G :=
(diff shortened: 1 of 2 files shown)


teorth:
Can the code be refactored using the new AllEquations.lean?

amirlb:
> Can the code be refactored using the new AllEquations.lean? I guess one should still record the list of constant and singleton equations somewhere, but not necessarily as a redundant list of defs

Done

teorth:
Congrats, this is the first approved automated contribution to the implication graph!  I've announced it on the front page and also made a blueprint chapter (should appear shortly as chapter 6).  If you could update that chapter to describe the process to generate the contributions when you have a chance, that would be great.

Command-Master:
@amirlb Could you please also PR the code used to produce these files?

amirlb:
> @amirlb Could you please also PR the code used to produce these files?

Where would be an appropriate place to put it? These files are still in the main `Generated/` directory, should I move them to a subdirectory and add a generating script there?

Issue #36, opened 2024-09-28
API: Create an API to handle Magma words, equations, and implications as syntactic objects


teorth:
See the discussion in https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/Syntax .  Among other things, this would allow one to prove "metatheorems" about equations, for instance that the equation "x = f(y,z,w)" is equivalent to "x=y" for any word f.

franklindyer:
claim

franklindyer:
propose PR #96

teorth:
See also https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/Equations.20vs.20Laws for further discussion

franklindyer:
propose PR #122

See my proof of `Equation332_implies_Equation43` for a proof-of-concept example of a duality argument. This is clearly still very awkward to write out, but hopefully with some more convenience functions like the ones in `FiniteSatUtils.lean`, this can get more pleasant to use.

I also have a question. For some reason, my proof doesn't type check with `(G: Type*)`, only with `(G: Type)`. I think this has something to do with how satisfaction is defined, or how my duality functions are defined, but I'm not clear on what the actual difference between `Type` and `Type*` is. Something about different type universes? Any idea how to fix this? And is there a reason we've chosen to write all of the implications in `Subgraph.lean` using `G: Type*` rather than `G: Type`?

teorth:
I opened a separate request #186 for this.  We've stated our implications to allow for the magma `G` to be "large" - not a member of the Level 0 universe, but potentially something from the Level 1 universe (e.g., a magma whose elements are arbitrary types in Level 0, and whose binary operation is, say, Cartesian product `X → Y → X × Y`).  But it is unlikely that we would actually encounter higher type objects in our arguments, so we could perhaps just reduce to the Type 0 universe.  On the other hand, the completeness theorem that is already proven should actually give us a way to pass back and forth between Level 0 and arbitrary level universes, so that may be the best long-term way to resolve the problem in case for some weird reason we do actually need to work with higher type objects at some point (and I think the Mathlib philosophy is to allow arbitrary level whenever possible).

Shreyas4991:
can we close this task closed given that we now have `FreeMagma` and `MagmaLaw` along with typical model theoretic properties lke soundness and completeness?

PR #45, opened 2024-09-28
Deep-embed graph

alxest:
Hi all! Thank you for initiating this extremely exciting project.
This is my attempt to simplify & give more principled approach to maintain the implication graph.

[Syntactic data]
Each axiom (equation) is now has a concrete, syntactic term, named `Equation` type. A graph, `Graph` type, is defined as `Equation -> Equation -> Status` where `Status` denotes the current state for the corresponding edge. Then, a concrete instance for the subgraph is given as a term `subgraph: Graph`.

[Semantic interpretation]
The validity of this graph object is enforced with `Graph.valid` definition - e.g., `subgraph_valid` would be the final theorem for the subgraph. The `Graph.valid` requires corresponding semantic proof for each syntactic edges. Notably, `Edge` is defined with typeclass and has few benefits below.

Benefits are as follows:
- In my experience, it is easier to play with and less error-prone to write a deep-embed the object (graph here) rather than doing meta-programming to generate it later, which typically relies on specific naming schemes on string (e.g., `EquationX_implies_EquationY` should have the exact type `(G: Type*) [Magma G] (h: EquationX G) : EquationY G`, which could be error-prone.)
- Simplification: (1) the repeated `(G: Type*) [Magma G]` for each theorem is now omitted, and (2) we don't have to give explicit names to theorems thanks to typeclass mechanism. (e.g., `instance : (Edge true .e2 .e4) where` instead of `theorem Equation2_implies_Equation4 (G: Type*) [Magma G] (h: Equation2 G) : Equation4 G :=`)
- Edge Synthesis: see `Edge.trans_true`, `Edge.abduct_false0` and `Edge.abduct_false1`: the typeclass mechanism allows us to nicely derive some facts from existing facts. This is tested in the [code](https://github.com/teorth/equational_theories/compare/main...alxest:equational_theories:deepembed-graph#diff-401dfd7eb752f8b8b969d15d6b990a698efe1fe63ea7d864282233b0a041693fR425-R430).

Surely, all these could be implemented manually, but it would be good to piggyback on existing language features in the start, until we reach the point that it doesn't scale. If you find this useful, I can make it a mergeable state quickly.

### DIFF
--- a/equational_theories/Equations.lean
+++ b/equational_theories/Equations.lean
@@ -2,44 +2,30 @@ import equational_theories.Magma
 import Mathlib.Tactic
 
 /- List of equational laws being studied -/
+inductive Equation: Type :=
+| e1 | e2 | e3 | e4 | e5 | e6 | e7 | e8 | e38 | e39
+| e40 | e41 | e42 | e43 | e46 | e168 | e387 | e4512 | e4513 | e4522 | e4582
 
-/-- The reflexive law -/
-def Equation1 (G: Type*) [Magma G] := ∀ x : G, x = x
-
-/-- The singleton law -/
-def Equation2 (G: Type*) [Magma G] := ∀ x y : G, x = y
-
-/-- The idempotence law -/
-def Equation3 (G: Type*) [Magma G] := ∀ x : G, x = x ∘ x
-
-/-- The left absorption law -/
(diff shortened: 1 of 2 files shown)


teorth:
I like this idea but it needs to be coordinated with #36 , perhaps one can discuss it on the Lean Zulip thread https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/Syntax ?

PR #46, opened 2024-09-28
proof of 46 implies 40

goens:
(had a proof_wanted)


Shreyas4991:
Hey, I'll approve this PR, but do please follow the PR claim process mentioned in the Outstanding tasks v2 thread. We don't want people to end up working in parallel and find their work not being used.

goens:
thanks! yes, sorry, I didn't see any task for this ine and it was small, but for anything meaningful I will do in the future!

PR #48, opened 2024-09-28
Introduce AllEquations.lean

nomeata:
as discussed in https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/Refactoring.20the.20Lean.20file.20structure/near/473253407


nomeata:
Unfortunately, the file is rather slow. So many operators doing the same type class inference over and over again!

```
cumulative profiling times:
	attribute application 139ms
	elaboration 15.8s
	fix level params 75ms
	import 65.6ms
	initialization 44.9ms
	instantiate metavars 29.4ms
	interpretation 2.7s
	linting 657ms
	parsing 1.72s
	process pre-definitions 628ms
	share common exprs 138ms
	type checking 346ms
	typeclass inference 85.6s
```

Maybe the `AllEquations` file could be rewritten to use `inst.op` explicitly, for faster processing. Or maybe it's fine as everyone builds this only once anyways.

PR #54, opened 2024-09-28
Trivial auto-generated theorems and tooling to compute transitive reductions/closures

vlad902:
This change includes ~15 thousand theorems, that prove a total of 4.2 million total derived implications (~20% of the total, e.g. 4694 ^ 2). I include scripts to calculate the transitive reduction and closure of a set of implications, to help future devs similarly submit minimal changes.


Command-Master:
Note that there is also `equational_theories/SimpleRewrites/src/find_redundant.py` for transitive reduction

vlad902:
I mentioned it on Zulip, but it should also go here, one thing I did not do here, is delete other theorems subsumed by these implications, e.g. in Subgraph.lean or other auto-generated implications. Should I? I'm not sure if that's kosher or desirable. Input is welcome.

vlad902:
> Note that there is also `equational_theories/SimpleRewrites/src/find_redundant.py` for transitive reduction

I am trying to run it for my base set of implications (1 million) and it hasn't finished after 5 minutes. (My script is also pretty slow tbf, it takes just under a minute.) I had to optimize my script so that it finished relatively quickly by performing condensing/uncondensing that I don't see here?

vlad902:
> I mentioned it on Zulip, but it should also go here, one thing I did not do here, is delete other theorems subsumed by these implications, e.g. in Subgraph.lean or other auto-generated implications. Should I? I'm not sure if that's kosher or desirable. Input is welcome.

e.g. this change subsumes `Generated/Singleton.lean`, but not `Generated/Constant.lean` or `SimpleRewrites`. In `Subgraph.lean`, it subsumes 43/46 implications.

teorth:
Thanks!  I just added a chapter to the blueprint at https://github.com/teorth/equational_theories/blob/main/blueprint/src/chapter/trivial_auto.tex , if at some point you could describe your methodology and compare the data set with other generated sets, that would be great.

PR #58, opened 2024-09-28
 Show there is some implication which is only falsified in an infinite magma

Command-Master:
I'm not sure I edited the blueprint correctly


teorth:
Looks good to me.  Only minor comments:

- One should add a \leanok to the proof as well as to the theorem, since you not only formalized the statements in Lean, but also provided proofs.
- Technically we are supposed to enclose each separate Lean contribution in their own namespace to avoid collisions, though in this case this is unlikely since we don't expect `EquationKis` to show up elsewhere.  (Though we could add it as Equation #4695, I guess, since it has some interest...)

teorth:
Ah, once you put the Lean implications in a namespace you have to adjust the blueprint \lean{} references accordingly.  I got caught out by this also.  I'll update CONTRIBUTING.md to warn about this.

Command-Master:
Could you expand on this? I was unable to resolve this without adding `import equational_theories.InfModel` to `equational_theories.lean`.

teorth:
Oh, right, you should do that also.

PR #68, opened 2024-09-28
Explain how to add many automatically generated proofs

carlini:
Per the discussion here, add a description of the suggested format when adding collections of large automatically generated proofs. https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/File.20structure.20for.20large.20auto-generated.20merges


vlad902:
Previously someone had mentioned putting auto-generated theorems under the `Generated/` directory and there seemed to be agreement, do we still want to do that? I'm not sure what's right

teorth:
I think that is a good idea.  Actually the latest autogenerated PR (simple_rewrites) was not placed inside generated, I'll ask the author to do that

carlini:
Updated the text to suggest under Generated/

Shreyas4991:
The instructions miss one step. The top level file of your project directory must be included in the top level file of the equational_theories folder. Without this step, the subproject won't be compiled

PR #81, opened 2024-09-28
add non-implications for equation 8

### LINKED ISSUE #30: SUBGRAPH: Add the implications/anti-implications between 8 and the other equations as proof_wanted statements.
Only a generating set of implications (and anti-implications) needs to be added to Subgraph.lean, of course.  Human-readable sketches of proofs can be added in comments.

edegeltje:
these non-implications should be sufficient to prove the relations to equation 8 complete.

Closes #30


edegeltje:
in case someone is surprised that `Equation40_not_implies_Equation8` was removed; it is redundant, since `46->40` and `not(46->8)`

PR #83, opened 2024-09-28
Code to find dual equations

### LINKED ISSUE #72: TOOL: code to compute the equation number of the dual of an equation
Example: Equation4564 is `x ∘ (y ∘ z) = (w ∘ y) ∘ z`.  Reversing all operations gives `(z ∘ y) ∘ = z ∘ (y ∘ w)`, which after relabeling and symmetry is Equation 4513, `x ∘ (y ∘ z) = (x ∘ y) ∘ w`.  Thus 4564 and 4513 are duals.  A code to output the dual of any given equation, and/or a table of equations and their duals, would be useful for the project.

carlini:
Fixes #72.

It might be nice to be able to also produce a proof of the duality of each of these equations in Lean. Right now you have to trust in my (and Claude's) coding ability. I believe it to be correct but that's not saying much.


teorth:
Once we have #36 we should be able to formalize duality in Lean, but for now a numerically generated list should be fine.  We can at least numerically test whether various numerically generated sets of implications are duality-invariant, which presumably they should be if they were generated by a duality-invariant scheme.

PR #87, opened 2024-09-28
Add a script that returns an equation ID from a written description

### LINKED ISSUE #79: TOOL: A tool that can take an equation as input, and return its equation number as output
For instance, `x ∘ (y ∘ z) = w ∘ (u ∘ z)` would return 4378.  Ideally, this tool could handle relabeling and symmetry (so one may first need to place the equation in a "normal form" in which the shorter expression is on the left, and variables are ordered suitably).  Could be useful for #72.

The equation numbering scheme is implicit in https://github.com/teorth/equational_theories/blob/main/scripts/generate_eqs_list.py .  The list is sorted by the total number of operations, then by the number of operations on the LHS. Within each such class we define an order on expressions by variable < operation, and lexical order on variables.

Ideally this tool would also work for equations with more than four operations.

carlini:
It works in all the cases I've tried. (All equations and relabelings of equations.) It tries to be somewhat smart with error handling to reject bad inputs. But it's not the most user-friendly tool.

Fixes #79.

### DIFF
--- a/scripts/find_equation_id.py
+++ b/scripts/find_equation_id.py
@@ -1,0 +1,200 @@
+import argparse
+import itertools
+from typing import List, Tuple, Union, Iterator
+
+EQ_SIZE = 4
+VAR_NAMES = 'xyzwuv'
+
+ExprType = Union[str, Tuple['ExprType', str, 'ExprType']]
+
+def tokenize(expr: str) -> List[str]:
+    """Convert an expression string into a list of tokens."""
+    expr = expr.replace(".", "∘").replace("(", " ( ").replace(")", " ) ").replace("∘", " ∘ ")
+    return [token for token in expr.split() if token]
+
+def parse_expr(tokens: List[str]) -> ExprType:
+    """Parse a list of tokens into an expression tree."""
+    def parse_element() -> ExprType:
+        if not tokens:
+            raise ValueError("Unexpected end of expression")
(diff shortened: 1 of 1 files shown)


carlini:
Test with, e.g., `python find_equation_id.py "(x . y) . z = (w .y) . w"` to get `The equation '(x . y) . z = (w .y) . w' is Equation 4648: ((x ∘ y) ∘ x) = ((z ∘ y) ∘ w)`

teorth:
Thanks!  Is your tool able to confirm the claim in https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/Infinite.20magmas.3F that the Kisielewicz law `(((y ∘ y) ∘ y) ∘ x) ∘ ((y ∘ y) ∘ z) = x` has number 374794 ?

carlini:
Yes, but out of the box the script assumes depth 4 with 6 variables so you'll have to modify the source to change that for now.

`The equation 'x = (((y ∘ y) ∘ y) ∘ x) ∘ ((y ∘ y) ∘ z)' is Equation 374794: x = ((((y ∘ y) ∘ y) ∘ x) ∘ ((y ∘ y) ∘ z))`

In running this though I've found a bug that will falsely claim `(((y ∘ y) ∘ y) ∘ x) ∘ ((y ∘ y) ∘ z) = x` (swapping on the equality) is not present.

PR #91, opened 2024-09-28
CONTRIBUTING: Verbiage to minimize automated implications by transitive reduction

vlad902:


Shreyas4991:
I recommend adding instructions about how to actually run and use the script. Perhaps you could automate the running with a CI action and post suggestions on any given PR as a list of duplicates?

Shreyas4991:
One possibility is to write a separate markdown file that explains the issue of duplicates and tiage them in detail and just link to it CONTRIBUTION.md

vlad902:
> One possibility is to write a separate markdown file that explains the issue of duplicates and how to use your scripts to triage them in detail and just link to it CONTRIBUTION.md

That's a good idea, I'll look at doing this tomorrow.

vlad902:
Sorry for the delay on this, the number of implementations of graph algos has seemingly doubled since I opened this change, maybe it makes sense to just mention how to get implications and tell people to ask on Zulip if they need further help? It seems like in ~2 days or so this will probably have a canonical solution.

Shreyas4991:
> Sorry for the delay on this, the number of implementations of graph algos has seemingly doubled since I opened this change, maybe it makes sense to just mention how to get implications and tell people to ask on Zulip if they need further help? It seems like in ~2 days or so this will probably have a canonical solution.

In that case it would be nice to have a separate zulip thread for this. The link to the topic should be in the contribution file to guide users to the right chat.

vlad902:
So, it's my fault for having let this languish, and I'm reviving this as this exact thing came up in #216 today. I realized was able to actually point them to the EquationSearch tool README since it already specifically calculates this reduction and includes a list of commands.

If you think it's still better to have an explicit runbook for automated theorem provers than the current link, I'd be happy to start that--I think I've been loathe to do so because I suspect I am probably the only person using my ruby tooling so I don't want to point to it as being the 'canonical' way; however having a link to "See how someone else solved this" seems less like setting the standard of The One True Way™.

teorth:
I think the perfect shouldn't be the enemy of the good.  By all means note your tool on CONTRIBUTING.md and in Zulip as one way to perform transitive reduction, but not claim that it is the canonical solution (though, if it works well enough for all other contributors, it may well become the defacto solution).

Shreyas4991:
I have a proposal: create a bulletted list of methods to use tools for graph reduction and a folder called `docs` that contains a markdown file explaining how to use your scripts.

vlad902:
I've updated it and left a TODO for reducing non-implications in `docs/graph_operations.md` as I haven't worked with it and I'm not sure the tooling. @nomeata Do you know if there is there scripts or any tooling to reduce non-implications I can point to somewhere?

PR #19, opened 2024-09-27
Massive generation of refutations via polynomials over Z/nZ

nomeata:
This PR adds a large number of automatically generated refutations by @carlini in https://github.com/carlini/equational_theories. All of them are magmas over Z/nZ with a polynomial of degree ≤2 as the operations.

For now, until we know better what to actually run, I excluded all magmas with 6 or more elements, just to keep this PR sizable.

PR #93, opened 2024-09-29
add `equational_result` attribute for tagging implications and negated implications

dwrensha:
Adds an attribute `@[equational_result]` to be used for tagging theorems that represent proven implications or nonimplications in the graph.

Collects these results in an environment extension. The idea is that anyone who wants to analyze the data about what's been proven can grab the data from that extension. The data includes the filename where the result lives, and can include any other metadata we want to include.

Previously we had discussed getting the same information by scanning through the environment and collecting all theorems that have the correct shape. That approach would have the advantage of allowing the theorems to be more succinct in the source code (no need to add an attribute). However, I prefer the approach taken by this PR, because:
* If a theorem does not quite match the correct shape, the attribute handler can report the error immediately, at the theorem declaration.
* Recording the results (via an environment extension) just once avoids the need to linearly scan the whole environment (could get expensive) at each downstream usage point.


Command-Master:
Shouldn't `ParseImplications` now be edited to use this instead of iterating over all theorems?

Command-Master:
Also, there is duplication of `Entry` and `EntryVariant` in `Conjecture` and `Result` which should be extracted to a common definition

dwrensha:
> Shouldn't ParseImplications now be edited to use this instead of iterating over all theorems?

Yes, it should. There are some details to be worked out about how exactly to organize the shared logic.

teorth:
Just a reminder to modify CONTRIBUTING.md to change the recommended syntax if this proposal gets implemented (it sounds reasonable to me so long as it doesn’t somehow break other Lean functionality concerning theorems)

Command-Master:
Using `result` instead of `theorem` should also be updated in other files

scripts/extract_implications.lean
```diff
@@ -1,4 +1,20 @@
...
-  println! (Lean.toJson (← generateOutput)).pretty
+import equational_theories.Result
+
+open Lean
+
+unsafe def main (args : List String) : IO Unit := do
+  let module := ← match args with
+    | [] => pure `equational_theories
+    | [mod] => pure mod.toName
+    | _ => throw <| IO.userError "usage: extract_impliations MODULE"
```
Command-Master, line 12: Perhaps you can use the CLI from #102

dwrensha:
> Shouldn't `ParseImplications` now be edited to use this instead of iterating over all theorems?

done.

dwrensha:
> Also, there is duplication of `Entry` and `EntryVariant` in `Conjecture` and `Result` which should be extracted to a common definition

The types do have differences and I don't see a benefit to trying to unify them.

It's possible that we may later decide to combine the `conjecture` and `equational_result` environment extensions. But for now I think this makes sense as-is.

scripts/extract_implications.lean
dwrensha: could make sense in a later PR. For now, I want to focus on getting the core functionality here landed.

Command-Master:
> > Also, there is duplication of `Entry` and `EntryVariant` in `Conjecture` and `Result` which should be extracted to a common definition
> 
> The types do have differences and I don't see a benefit to trying to unify them.
> 
> It's possible that we may later decide to combine the `conjecture` and `equational_result` environment extensions. But for now I think this makes sense as-is.

What are the differences? To me they seem like the same thing, except one is missing a proof. The benefit of merging them is that later code, e.g. transitive closure, could use the same interface to process both of them

equational_theories/EquationalResult.lean
```diff
@@ -0,0 +1,103 @@
...
+
+namespace Result
+
+/-- An entry in the equational results environment extension.
+-/
+inductive EntryVariant where
+  | implication : Implication → EntryVariant
+  | nonimplication : Implication → EntryVariant
+  /-- An equation that always holds. -/
+  | unconditional : String → EntryVariant
```
nomeata, line 16: These two are just instance of the general “∃ magma G such that equations {1,2,3} hold and {4,5,6} do not hold”. Note that that already entails 9 non-implicaitons. Once we add more exhaustive analysis of example magmas (as in #19), listing all non-implications here will not scale, and we’ll probably have to switch to the more compact representation.

Doesn't have to happen in this PR, of course; this is just a heads up.

equational_theories/EquationalResult.lean
```diff
@@ -0,0 +1,103 @@
...
+syntax (name := printEquationalResults) "#print_equational_results" : command
+
+elab_rules : command
+| `(command| #print_equational_results) => do
+  let rs ← extractEquationalResults
+  for ⟨name, _filename, res⟩ in rs do
+    match res with
+    | .implication ⟨lhs, rhs⟩ => println! "{name}: {lhs} → {rhs}"
+    | .nonimplication ⟨lhs, rhs⟩ => println! "{name}: ¬ ({lhs} → {rhs})"
+    | .unconditional rhs => println! "{name}: {rhs} holds unconitionally"
```
nomeata, line 83:
```suggestion
    | .unconditional rhs => println! "{name}: {rhs} holds unconditionally"
```

equational_theories/EquationalResult.lean
```diff
@@ -0,0 +1,103 @@
...
+  let rs := equationalResultsExtension.getState (← getEnv)
+  let mut implications : List Implication := []
+  let mut nonimplications : List Implication := []
+  let mut unconditionals : List String := []
+  for ⟨_name, _filename, res⟩ in rs do
+    match res with
+    | .implication imp => implications := imp::implications
+    | .nonimplication nimp => nonimplications := nimp::nonimplications
+    | .unconditional s => unconditionals := s::unconditionals
+  return ⟨implications, nonimplications, unconditionals⟩
```
nomeata, line 103: Better to use `Array` here, we expect these to be maybe not neglectible size.

dwrensha: done in https://github.com/teorth/equational_theories/pull/93/commits/a4d113eef09a2db84f5bf19aee4fceced68e2d17

equational_theories/EquationalResult.lean
dwrensha: done in https://github.com/teorth/equational_theories/pull/93/commits/a4d113eef09a2db84f5bf19aee4fceced68e2d17

equational_theories/EquationalResult.lean
dwrensha: I agree that we will want to add other ways to represent results. That can happen later as needed.

I don't see how "EquationN holds for all magmas" is an instance of "∃ magma G such that equations {1,2,3} hold and {4,5,6} do not hold".

dwrensha:
> > > Also, there is duplication of `Entry` and `EntryVariant` in `Conjecture` and `Result` which should be extracted to a common definition
> > 
> > 
> > The types do have differences and I don't see a benefit to trying to unify them.
> > It's possible that we may later decide to combine the `conjecture` and `equational_result` environment extensions. But for now I think this makes sense as-is.
> 
> What are the differences? To me they seem like the same thing, except one is missing a proof. The benefit of merging them is that later code, e.g. transitive closure, could use the same interface to process both of them

`conjecture` currently has an `other` variant because I wanted to support the ability to make conjectures whose types don't fit the enumerated shapes.

I could imagine that if we add new variants to support e.g. compactly representing many negated implications at once, that such things might only make sense as fully baked results, not as conjectures.

I could see it making sense to merge these into one environment extension or to keep them separate. Keeping them separate seems like the less invasive thing right now. We can adapt in the future. I think more important than nailing down this backend interface is putting the frontend in place, i.e. the `conjecture` command and the `equational_result` attribute, so that everyone else working on the project can start using them.

equational_theories/EquationalResult.lean
nomeata: A, I misread the comment. My bad

PR #94, opened 2024-09-29
Simple chaining

kant2002:
This is another trivial thing, in comparison how you guys think about the problem. I see `Equation46_implies_Equation387` and `Equation387_implies_Equation43` and as such I can produce `Equation46_implies_Equation43`.

This is lead me to thinking that maybe this kind of reasoning can be automated.


kant2002:
Okay. Found `scripts\transitive_reduction_exact.rb` which seems to be do exactly thing I think about.

teorth:
Our current thinking is to try to minimize the number of implications that are explicitly in the Lean files to keep the compile times fast, and rely on post processing such as the transitive reduction to complete the implication graph afterwards.

PR #96, opened 2024-09-29
added an initial attempt at treating magma words as syntactic objects to facilitate proving metatheorems

franklindyer:
The inductive type `FreeMagma` implements abstract syntax trees over the magma operation (i.e. just binary trees), with the leaves indexed by a given type. This lets us express magma words like `(y ∘ z) ∘ w)` as tree data structures `(Lf 0 ⋆ Lf 1) ⋆ Lf 2`. The function `evalInMagma` evaluates the syntax tree in a certain magma of choice, for a given specialization of the leaf values.

Included is an example of a "metatheorem" saying roughly that whenever a universal identity of the form `x0 = f(x1,x2,...)` holds, then the universal identity `x = y` holds. It is applied to prove both `Equation37 -> Equation2` and `Equation514 -> Equation2`.

This approach is a bit wanting for syntactic sugar, but it could be a start!

### DIFF
--- a/equational_theories/FreeMagma.lean
+++ b/equational_theories/FreeMagma.lean
@@ -1,0 +1,43 @@
+import equational_theories.Conjecture
+import equational_theories.AllEquations
+
+universe u
+universe v
+
+inductive FreeMagma (α : Type u)
+  | Leaf : α → FreeMagma α
+  | Fork : FreeMagma α → FreeMagma α → FreeMagma α
+
+infixl:65 " ⋆ " => FreeMagma.Fork
+def Lf {α : Type u} : (α → FreeMagma α) := FreeMagma.Leaf
+
+def fmapFreeMagma {α : Type u} {β : Type v} (f : α → β) : FreeMagma α → FreeMagma β
+  | FreeMagma.Leaf a => FreeMagma.Leaf (f a)
+  | FreeMagma.Fork lchild rchild => FreeMagma.Fork (fmapFreeMagma f lchild) (fmapFreeMagma f rchild)
+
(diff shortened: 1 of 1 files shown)


teorth:
I like the use of free Magmas here.

Perhaps one could encode the syntactic version of an equation law as an element `E` of`FreeMagma ℕ × FreeMagma ℕ`, with a magma `G` obeying the law if `∀ f : ℕ → G, evalInMagma f E.1 = evalInMagma f E.2` (one could call this something like `G ⊧ E` perhaps).  The type of such laws should then come with a `PreOrder` for implication, with the syntactic version of Equation1 and Equation2 being `⊥` and `⊤` , and there should also be a duality operation.

Not sure how best to integrate the syntactic laws with the semantic ones that we currently have.  Maybe each `EquationX` is going to come with a syntactic counterpart `SynEquationX`, perhaps, accompanied with some easy (maybe even `rfl`) lemma to the effect that `∀ (G:Type) (hG: Magma G), EquationX G ↔ (G ⊧ SynEquationX)`.  Perhaps there are other suggestions though...

codyroux:
I suspect there will be soundness and completeness lemmas, and then passing from syntax to semantics is simply an `eval` in the `->` direction and (unfortunately) a tactic or macro in the `<-` direction, which simply turns an equality back into its syntax as needed.

Not sure there's a need to generate lemmas.

IIRC the completeness theorem for equational logic is quite a bit easier than for general structures, it might be worthwhile to re-prove it ([Birkhoff's theorem](https://en.wikipedia.org/wiki/Birkhoff%27s_theorem_(equational_logic)).

PR #97, opened 2024-09-29
Delete transitive_reduction_exact.rb

vlad902:
I did not intend to upstream this script, it is too slow and it is confusing to have two transitive reduction scripts.


pitmonticone:
Please merge from `main` and the CI checks will pass.

I have just fixed a mistaken blueprint declaration.

PR #99, opened 2024-09-29
transitive_reduction: Fix SCC computation bug

vlad902:
Due to a silly bug (index being a local instead of a global variable here), the SCC computation under-computed SCCs. The previous transitive reduction results were still correct, but slower to compute than necessary.


pitmonticone:
Please merge from main and the CI checks will pass.

PR #100, opened 2024-09-29
Cache build artefacts in blueprint workflow

pitmonticone:


.github/workflows/blueprint.yml
```diff
@@ -48,29 +48,23 @@ jobs:
...
-            .lake/build/doc/Init
-            .lake/build/doc/Lake
-            .lake/build/doc/Lean
-            .lake/build/doc/Std
-            .lake/build/doc/Mathlib
-            .lake/build/doc/declarations
-            .lake/build/doc/find
-            .lake/build/doc/*.*
+            .lake/build
+            !.lake/build/doc-data
```
nomeata, line 56: I'd say we should exclude the mathlib directory therein as well, it's very large and probably faster to fetch via `lake exec cache get`

.github/workflows/blueprint.yml
```diff
@@ -48,29 +48,23 @@ jobs:
...
             !.lake/build/doc/declarations/declaration-data-equational_theories*
-          key: MathlibDoc-${{ hashFiles('lake-manifest.json') }}
-          restore-keys: MathlibDoc-
+          key: LakeBuild-${{ hashFiles('lake-manifest.json') }}
+          restore-keys: LakeBuild-
```
nomeata, line 59: GitHub action caching is very confusing here, but this will not work: the `key` needs to include the git ref, else it will never push a new cache if the lake manifest stays the same. 

Maybe follow this example:
https://github.com/leanprover/lean-action/blob/9e84bf82a65182a555b9c9b4b78500fa203dc5a4/action.yml#L136

.github/workflows/blueprint.yml
```diff
@@ -79,7 +73,6 @@ jobs:
           else
             echo "The 'home_page' folder does not exist in the repository."
             echo "HOME_PAGE_EXISTS=false" >> $GITHUB_ENV
-          fi
```
nomeata, line 76: Intentional?

pitmonticone: Nope, I'll fix it.

.github/workflows/blueprint.yml
pitmonticone: Done. Thanks.

.github/workflows/blueprint.yml
pitmonticone: Which is the exact path you're referring to? `.lake/build/doc/Mathlib`?

nomeata: At the phone right now. I think the path is something like `.lake/packages/mathlib` or so - look at your local `.lake` directory :-)

`batteries` can also be excluded.

pitmonticone: But those are already excluded, right? They're not within the `.lake/build` directory so it's not supposed to be cached by this configuration.

nomeata: Ah, I missed that! Nevermind then.

(I wonder if we should include libraries we depend upon that are not mathlib or dependend upon by mathlib, like egg. But we can iterate!)

pitmonticone: Ok, I would merge for now.

PR #110, opened 2024-09-29
Example file with “massive” couterexample, with infrastructure

nomeata:
this extracts the “infrastructure” built in #19 into its own PR. In
particular

 * The `Facts G [1,2] [3,4]` syntax to concisely and efficiently state
   that G satisfies Equation1 and Equation2, but not Equation3 and
   Equation4

   There is also a crude `calculate_facts` command that takes a magma
   and an instance and a tactic and lists those for which the tactic
   succeeds.

* The `decideFin!` tactic that is a bit like `decide` but simply
  believes that the statement is true, so that the check only happens by
  the kernel, and uses custom code to construct the `Decidable` instance
  for the kind of theorems we actually encounter, to avoid the (slow)
  type inference.

* The `memoFinOp` elaborator that take a function
  `Fin n → Fin n → Fin n`, tabulates it, encodes this table in a single
  number and implements the lookup based on that. Presumably that's more
  compact and faster to evaluate by `decide` than other forms (array
  indexing, polynomials etc.)


equational_theories/DecideBang.lean
```diff
@@ -0,0 +1,87 @@
...
+private def splitConjs (e : Lean.Expr) : Array (Lean.Expr) := Id.run do
+  let mut e := e
+  let mut r := #[]
+  while e.isAppOf `And do
+    r := r.push e.appFn!.appArg!
+    e := e.appArg!
+  r := r.push e
+  return r
+
+elab "decide!" : tactic => do
```
eric-wieser, line 30: Can you document this one too?

equational_theories/MemoFinOp.lean
```diff
@@ -0,0 +1,67 @@
...
+
+open Lean Meta Elab  Term
+
+def opOfTable {n : Nat} (t : Nat) (a : Fin n) (b : Fin n) : Fin n :=
+  let i := a.val * n + b.val
+  let r := (t / n^i) % n
+  ⟨r, Nat.mod_lt _ (by exact Fin.pos a)⟩
+
+/-- `enum n` is the array of all elements of `Fin n` in order -/
+private def Fin.enum (n) : Array (Fin n) := Array.ofFn id
```
eric-wieser, line 17:
```suggestion
private def Array.finRange (n) : Array (Fin n) := Array.ofFn id
```
to match `List.finRange`

equational_theories/MemoFinOp.lean
```diff
@@ -0,0 +1,67 @@
...
+  t
+
+-- Quick sanity check:
+example :
+    let f := fun (a b : Fin 4) => a * a * b + 2
+    opOfTable (buildMemo f) = f := by
+  funext a b; revert a b; native_decide
+
+@[implemented_by Meta.evalExpr]
+private opaque myEvalExpr (α) (expectedType : Expr) (value : Expr) (safety := DefinitionSafety.safe) : MetaM α
```
eric-wieser, line 34: What's the purpose of this? Can you use `unsafe evalExpr ..` at the caller instead

equational_theories/MemoFinOp.lean
```diff
@@ -0,0 +1,67 @@
+import Lean
+
+/-!
+This file defines a macro `memoFinOp` that takes a function `f : Fin n → Fin n → Fin n`, at
+elaboration time calculates its table of values, encodes it into a single `Nat`, and returns a term
```
eric-wieser, line 5: Is the trick here that nat arithmetic "lookups" are faster than storing a flat array? Is kernel reduction speed the relevant performance metric here, or runtime evaluation?

equational_theories/MemoFinOp.lean
nomeata: It's a copy of Fin.enum from Batteries.Data.Fin.Basic. And it's private… :-)

equational_theories/DecideBang.lean
nomeata: 👍🏻 Will do tomorrow

equational_theories/MemoFinOp.lean
nomeata: There is no `unsafe elab`, so some `implementedBy` is needed. I can be less sloppy and make the elaboration function body a safe wrapper or so, though.

equational_theories/MemoFinOp.lean
nomeata: Yes, in the kernel there are no arrays, they become lists, and then lookup is likely quite slow.

I think kernel reduction is the right metric, as long as we use `by decide` and not `by native_decide`

equational_theories/MemoFinOp.lean
eric-wieser: But there is an `unsafe` `elab`, in that there is an elaborator `unsafe $t:term : term` which makes the auxiliary definition for you.

nomeata: Oh, I thought I tried that. Will try again tomorrow. (Is it maybe a mathlibism?)

equational_theories/MemoFinOp.lean
eric-wieser: Ah, I mistakenly thought leanprover-community/batteries#899 had been merged.

equational_theories/MemoFinOp.lean
Command-Master: Wasn't the conclusion from Zulip that it's better to use a base which is a power of two, with shifts and bitwise and to extract the data?

PR #113, opened 2024-09-29
add DSL for definging magma-laws

edegeltje:
this PR introduces syntax and elaboration for defining terms of `Prod (FreeMagma String) (FreeMagma String)`.


codyroux:
I think this clashes with https://github.com/teorth/equational_theories/blob/main/equational_theories/MagmaLaw.lean, and is perhaps subsumed by it?

PR #116, opened 2024-09-29
Prove some implications taken from old Putnam problems

franklindyer:
Proved a couple of specific implications drawn from old Putnam problems:

- Putnam 1978, Problem A4 parts (a) and (b)
- Putnam 2001, Problem A1


teorth:
Putnam questions seem to have been easier in the past :).

PR #122, opened 2024-09-30
utilities to be used for writing metatheorems about duality

franklindyer:
For now, this is just a start.

The most notable metatheorem here so far is `DualizeEquationLawImplication`, which allows us to convert `e1 = e2 -> e3 = e4` to `e1* = e2* -> e3* = e4*`. However, it is still very clunky to use, since `⊧` is defined in terms of substitution functions `Nat -> G`. As a proof of concept I'd like to add an example showing how `DualizeEquationLawImplication` can be used to deduce `Equation3_implies_Equation23` from `Equation3_implies_Equation8`, but I haven't gotten to this yet.

I also notice that @codyroux has given their own definition for `⊧` which differs from mine in that it is more polymorphic. This is fine - I think their PR should be merged before this one, and then I'll clobber my definition with theirs and fix any of my existing theorems to work with that definition.


codyroux:
Cool! The PR is now #123, which contains more stuff (and hopefully passes the dry-run). I'm happy to fix this myself if it's easier.

franklindyer:
@codyroux I've gone ahead and updated my duality work to be consistent with your definition of `⊧`. (I like your polymorphic version better. ☺️)

franklindyer:
awaiting-review

This is ready for review now that #123 is merged.

codyroux:
It seems that this is overlapping with https://github.com/teorth/equational_theories/blob/main/equational_theories/MagmaOp.lean, sadly we should have coordinated earlier. Maybe this one can subsume that one? @franklindyer do you want to open a channel on the Zulip to discuss this?

franklindyer:
> It seems that this is overlapping with https://github.com/teorth/equational_theories/blob/main/equational_theories/MagmaOp.lean, sadly we should have coordinated earlier. Maybe this one can subsume that one? @franklindyer do you want to open a channel on the Zulip to discuss this?

Oops, I did not notice that. Looks like it proves the duality theorem a heck of a lot more quickly haha - I will just go ahead and close this PR. Might open separate PRs for some of bits and pieces that can be salvaged from this.

PR #123, opened 2024-09-30
Codyroux/compactness

### LINKED ISSUE #104: Build the basic meta-theory for the (syntactic) equational theory of magmas
~~No need to bring proofs into this (yet)~~, but this should enable proving compactness, and the meta-theorems alluded to [here](https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/Syntax/near/473455037)

codyroux:
closes #104 

Closes #104


equational_theories/Completeness.lean
```diff
@@ -0,0 +1,220 @@
...
+  ∀ (G : Type)[Magma G], satisfiesSet G Γ → satisfies G E
+
+infix:50 " ⊧ " => (satisfies)
+
+infix:50 " ⊧ " => (satisfiesSet)
+
+infix:50 " ⊧ " => (models)
+
+infix:50 " ⊢ " => (derive)
+
```
franklindyer, line 73: Maybe the definitions up to this point can ultimately end up in their own file, called something like `Satisfaction.lean`? I suggest this because I expect lots of other files (like my `Duality.lean` in progress) will need to use these definitions, but may not need everything in `Completeness.lean`.

codyroux:
I agree, I'll either do it here, or in a further PR.

equational_theories/Completeness.lean
codyroux: I agree! I'd like to get this merged first, if that's all right, I'm getting nervous about all the rebasing.

codyroux:
awaiting-review

pitmonticone:
Review here: https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/Bluprint.20help/near/473837219

nomeata:
Was this meant to add a file called `Unidecode-1.3.8-py3-none-any.whl`?

PR #127, opened 2024-09-30
Add `@[equational_result]` attribute to generated results

Command-Master:


Command-Master:
I see I somehow missed many files, I'll fix this ASAP

equational_theories/AllEquations.lean
```diff
@@ -1,6 +1,7 @@
 import equational_theories.Magma
 import equational_theories.Equations
 import equational_theories.EquationsCommand
+import equational_theories.EquationalResult
```
nomeata, line 4: That doesn’t seem to be needed for this file, does it? Generally maybe better to add it in the files that actually need this, to keep the module graph more parallel, and that you don’t have to recompile `AllEquations` just because the attribute code changes.

nomeata:
I’m a bit confused – if these are generated files, shouldn’t there be a script that generates them, and shouldn’t you update (and rerun) that script?

Command-Master:
> I’m a bit confused – if these are generated files, shouldn’t there be a script that generates them, and shouldn’t you update (and rerun) that script?

I could update the scripts, but I don't see the point in rerunning them - it won't change what they produce

nomeata:
> I could update the scripts, but I don't see the point in rerunning them - it won't change what they produce

It’s really important that the generated files can always be re-generated from the scripts. Both to do science in a reproducible and transparent way, but also if we need to make more substantial refactorings that you really don’t want to do by editing huge files.

So I strongly suggest to *never* edit generated files, and always update them by adjusting the scripts. Ideally the scripts leave a comment in the file to that effect.

Command-Master:
I agree, although that might cause a problem once we start to have scripts which take a very long time to run. Hopefully there won't be such changes by them. Where's the code for `Constant.lean`, `SimpleRewrites.lean`, `Singleton.lean`, `TrivialBruteforce.lean` and `Z3GenNonImp.lean`?

nomeata:
> I agree, although that might cause a problem once we start to have scripts which take a very long time to run.

In that case ideally the scripts return a txt or json file with the “raw data”, and a separate script translates to Lean. This is the idiom that #111 and #19 follow.

> Hopefully there won't be such changes by them. Where's the code for Constant.lean, Singleton.lean?

I was wondering about that as well 😨

equational_theories/AllEquations.lean
nomeata: It was quite slow to compile `AllEquations`, but maybe it got faster. 🤷🏻

nomeata:
Anyways, looks good and I want that data now, so I’ll merge :-)

Issue #132, opened 2024-09-30
API: Give MagmaLaw the structure of a Preorder (Lemma 3.2)


teorth:
Thus, if `E`, `E'` are laws, then `E ≥ E'` will be true precisely when `E` implies `E'`, that is to say every magma that obeys `E` obeys `E'`.  The axioms of a preorder will need to be verified.

When done, mark Lemma 3.2 in the blueprint with a \leanok in both the statement and proof.

goens:
claim

goens:
propose PR#135

goens:
awaiting-review

PR #135, opened 2024-09-30
feat: add preorder on magma laws

goens:
This builds on the `MagmaLaw` structure defined in #113 and thus depends on it.


goens:
awaiting-review

goens:
This seems to conflict with #123 (which also defines `MagmaLaw`), can also refactor it to use those definitions once that's merged.

teorth:
It seems that #123 is merged now, so what is the status on compatibility?

codyroux:
It might be worth defining `l ≤ l'` as `{l} ⊧ l'` while we're at it, but it's not critical.

goens:
good idea @codyroux, thanks, I just added that as an equivalent characterization (`l₁ ≤ l₂ ↔ {l₁} ⊧ l₂`), thought the other definition was simpler without having to deal with the singleton

It's integrated with #132 now, I'm happy to merge this if you're happy (@teorth)

equational_theories/FreeMagma.lean
```diff
@@ -21,16 +21,20 @@ notation "Lf" => FreeMagma.Leaf
 
 instance FreeMagma.Magma {α} : Magma (FreeMagma α) := ⟨ Fork ⟩
```
madvorak, line 22: I would call it `instance FreeMagma.isMagma` instead.
Then you don't need the `_root_` thing.

equational_theories/Completeness.lean
```diff
@@ -24,6 +26,8 @@ def Ctx α := Set (MagmaLaw α)
 -- FIXME: figure out how to remove this.
 instance Ctx.Membership α : Membership (MagmaLaw α) (Ctx α) := ⟨ Set.instMembership.mem ⟩
 
+instance {α : Type} : Singleton  (MagmaLaw α) (Ctx α) := ⟨Set.singleton⟩
```
madvorak, line 29: double space

equational_theories/Preorder.lean
```diff
@@ -0,0 +1,36 @@
...
+We have to explicitly quantify the type `G` and the Magma instance `[Magma G]` instead of
+using them as parameters so that the implication holds in any Magma `G`.
+-/
+def implies (l₁ l₂ : MagmaLaw α) := ∀ {G : Type} [Magma G],
+  satisfies G l₁ → satisfies G l₂
+
+instance : LE (MagmaLaw α) where
+  le l₁ l₂ := l₁.implies l₂
+
+theorem implies_refl (l : MagmaLaw α) : l ≤ l := by
```
madvorak, line 21: This should work afaik:
```
theorem implies_refl (l : MagmaLaw α) : l ≤ l :=
  fun _ _ => rfl
```

equational_theories/FreeMagma.lean
```diff
@@ -21,16 +21,20 @@ notation "Lf" => FreeMagma.Leaf
...
-def evalInMagma {α : Type u} {G : Type v} [Magma G] (f : α -> G) : FreeMagma α → G
+def evalInMagma {α : Type u} {G : Type v} [_root_.Magma G] (f : α -> G) : FreeMagma α → G
   | Lf a => f a
   | lchild ⋆ rchild => (evalInMagma f lchild) ∘ (evalInMagma f rchild)
 
+end FreeMagma
+
 theorem ExpressionEqualsAnything_implies_Equation2 (G: Type u) [Magma G]
-  : (∃ n : Nat, ∃ expr : FreeMagma (Fin n), ∀ x : G, ∀ sub : Fin n → G, x = evalInMagma sub expr) → Equation2 G := by
+  : (∃ n : Nat, ∃ expr : FreeMagma (Fin n), ∀ x : G, ∀ sub : Fin n → G, x = expr.evalInMagma sub) → Equation2 G := by
```
madvorak, line 37: When we are at it, colon goes at the end of the previous line.

equational_theories/Preorder.lean
goens: nope, it doesn't:
```
type mismatch
  rfl
has type
  ?m.104 = ?m.104 : Prop
but is expected to have type
  satisfiesPhi x✝ l : Prop
  ```
  but even if it did, what's wrong with the current proof? looks perfectly reasonable to me.

goens:
thanks for the review @madvorak, I've made all the style changes you suggested (except the proof change that didn't work)

equational_theories/FreeMagma.lean
madvorak: Sorry for the misleading comment.

equational_theories/Preorder.lean
madvorak: Sorry for the misleading comment.

madvorak (approved):

equational_theories/FreeMagma.lean
codyroux: That's better for sure!

pitmonticone:
Please let me know if there are any further review comments @codyroux. If there are not, I will merge it so we can merge #165 too.

pitmonticone:
Thanks everyone! Merged.

PR #137, opened 2024-09-30
Type*

madvorak:
If we use `Type*`, we will not have the schisma between `universe u` and variable `u`.

I also made a tiny change to misplaced whitespaces in `Subgraph.lean`.


madvorak:
In this case, do we want to change the naming convention for variables or the names of universes?

madvorak:
FYI `Mathlib.Tactic.TypeStar` imports next to nothing.

nomeata:
> In this case, do we want to change the naming convention for variables or the names of universes?

The universe, I’d say. Would using `Type _` not just work fine?

> FYI Mathlib.Tactic.TypeStar imports next to nothing.

It’s still something :-). Maybe it matter less, but it’s nice for example to be able to copy things to live.lean-lang.org for bug reports etc.

PR #141, opened 2024-09-30
easy refactoring of a few unnecessarily complicated proofs

madvorak:


madvorak:
no need to call `linarith` for `0 = 1`

PR #144, opened 2024-09-30
Update scripts/process_implications.py: use Lean parser, output irreducible implications

amirlb:
Update scripts/process_implications.py: to process the parsed implications JSON, and figure out the irreducible implications

### DIFF
--- a/scripts/process_implications.py
+++ b/scripts/process_implications.py
@@ -1,28 +1,32 @@
-#! /usr/bin/env python3
+#! /usr/bin/env python
 
 """
 Example usage:
 
 ```sh
-$ python scripts/process_implications.py equational_theories/Subgraph.lean
+$ lake exe extract_implications | python scripts/process_implications.py
 ```
 """
 
 from collections import defaultdict
+import json
 import os
 from random import sample
 import re
-from sys import argv
+from sys import argv, stdin
+import networkx as nx
 
 
 def transitive_closure(pairs):
+    pairs_idx = defaultdict(list)
+    for a, b in pairs:
+        pairs_idx[a].append(b)
     new_pairs = closure = set(pairs)
     while new_pairs:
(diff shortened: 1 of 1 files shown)


vlad902:
Running this gave me `All implications: 310354` while my own output is millions higher, am I misunderstanding the output or is that the size of the entire closure?

vlad902:
```
lake exe extract_implications | ruby -rjson -e 'JSON.parse($stdin.read)["implications"].each { |s| puts s["lhs"][8,10] + "," + s["rhs"][8,10] }' | sort -u > /tmp/imp.csv
ruby scripts/transitive_closure.rb  /tmp/imp.csv | wc -l
```

Gives me 4.5 million, but I would totally believe the error is in `scripts/transitive_closure.rb`

amirlb:
The figure 310354 is after reducing to the 3557 equivalence classes, without the internal implications in each class and with a single implication between every two related classes. That's probably the source of the difference.

teorth:
Would it be possible to update the README.md to give links to this data and script, possibly with some warning that the data set is provisional and not guaranteed to be 100% accurate yet?  Having a easily accessible initial data set of implications, even if slightly off from the actual truth, may allow a lot of more data analysis / data visualization oriented people to be able to make some interesting contributions and insights.  (One can also think about adding a chapter to the blueprint to give a longer description of the methodology.)

vlad902:
> The figure 310354 is after reducing to the 3557 equivalence classes, without the internal implications in each class and with a single implication between every two related classes. That's probably the source of the difference.

Ah, that makes sense. Given that there's two different graphs to be considering here, pre- and post-condensation, it's probably good to be explicit about it. I'd maybe say 'condensed implications' or something else you find more descriptive to make it clear to others who may be confused like me.

PR #150, opened 2024-09-30
Brute-force enumeration over all Magma 4x4 [corrected]

carlini:
This is a corrected rebase of #111, after I incorrectly rebased and added some wrong files.

Description from there:

This branch makes partial progress towards refuting anything that can be done with a 4x4 magma. There are 4^(4*4) = 2^32 of them, and each needs to be checked on all 4696 equations. This C code can brute force this in a few hundred CPU hours. It could be made much more efficient if needed (likely at least 10-100x) by keeping everything in memory and not using the disk so much. Also, it is not thread safe and has race conditions. But I've run it several times and got the same output each time so I'm confident in these results being complete.

The final output is a bunch of lines of the form

Table [[2, 2, 0, 3], [2, 2, 1, 0], [2, 2, 0, 3], [2, 3, 1, 0]]
Proves [1, 411, 416, 4065, 4090, 4131, 4590, 4622]

Which you should interpret as meaning that this 4x4 magma is true only on these equations, and false on all others.

Inspired by @euprunin in https://github.com/teorth/equational_theories/pull/111#issuecomment-2382246989 I've tried to show a step-by-step proof. So you can interpret this as follows, which the code will now also walk through the proof of:

Showing that def Equation411 (G: Type*) [Magma G] := ∀ x : G, x = x ∘ (x ∘ (x ∘ (x ∘ x)))
is correct on Magma
[[2 2 0 3]
 [2 2 1 0]
 [2 2 0 3]
 [2 3 1 0]]
Checking equation on assignments x=0
table[0, 0] = 2
table[0, 2] = 0
table[0, 0] = 2
table[0, 2] = 0
Checking equation on assignments x=1
table[1, 1] = 2
table[1, 2] = 1
table[1, 1] = 2
table[1, 2] = 1
Checking equation on assignments x=2
table[2, 2] = 0
table[2, 0] = 2
table[2, 2] = 0
table[2, 0] = 2
Checking equation on assignments x=3
table[3, 3] = 0
table[3, 0] = 2
table[3, 2] = 1
table[3, 1] = 3

As suggested by @teorth in https://github.com/teorth/equational_theories/pull/111#issuecomment-2381986345 this also updates the README and adds some placeholder to the blueprint


carlini:
Perfect. Let me know if you have other questions and want to use this for anything.

PR #151, opened 2024-09-30
no associativity -> better in infoview

madvorak:


madvorak:
Infoview will now display all parentheses.

PR #152, opened 2024-09-30
Codyroux/derivation preserves size

codyroux:
[DO NOT MERGE] Use completeness to prove a simple syntactic meta-theorem about the size of derived equations.

Basically the only equations derivable from a set of equalities `E` are either "trivial", that is of the form `t = t` or the size of each term is larger than the minimum of the terms found in `E`.


teorth:
Sorry I had to revert your PR as `OrderMetatheorem.lean` stopped compiling for some reason, you may need to resubmit it

codyroux:
No biggie, I'll re-create

Issue #114, opened 2024-09-29
API: Develop an API for homomorphisms and isomorphisms between magmas


teorth:
In exact analogy to how Mathlib supports additive homomorphisms (`AddHom`) `φ: G →+ H` between `(G:Type) [Add G]` and `(H:Type) [Add H]`, and multiplicative homomorphisms (`MulHom`) `φ: G →* H` between `(G:Type) [Mul G]` and `(H:Type) [Mul H]`, expand `Magma.lean` to support magma homomorphisms (`MagmaHom`) `φ: G →∘ H` between `(G:Type) [Magma G]` and `(H:Type) [Magma H]`, with all the basic API copied over from Mathlib.  Similarly create support for `MagmaEquiv` analogously to `AddEquiv` and `MulEquiv`. This will help support #104 and #36.

madvorak:
claim

madvorak:
https://github.com/teorth/equational_theories/pull/158

madvorak:
propose PR #158

PR #155, opened 2024-10-01
Free magma universal properties

### LINKED ISSUE #133: API: Show that the free magma has the universal property (Definition 1.4)
Given a map `f : α -> G`, show that `evalInMagma f` is the unique magma homomorphism that extends `f`.  (This requires that #114 is completed.  Additionally, `evalInMagma f` should now have the type `MagmaHom`.)

Similarly, if `f : α → β`, show that `fmapFreeMagma f` is the unique magma homomorphism that extends `f`.  (One can add some API relating these two operations as necessary.)

When done, mark Definition 1.4 with a \leanok on the blueprint.

franklindyer:
Added univeral properties of `evalInMagma` and `fmapFreeMagma`. (The universal property of the latter follows directly from that of the former!)

Closes #133.

I've written my own `isMagmaHom` predicate for the sake of this issue, but it will need to be replaced with the "official" `MagmaHom` implementation from #114 once it is complete. So this PR should not be merged until that issue is resolved, and the necessary tweaks are made.


Shreyas4991:
@franklindyer : Please claim an issue as described in the contribution.md file, before linking a PR. Not doing so messes up the task categorisation.

teorth:
#158 is in now, so if one can refactor this PR to use that API, that would be great.

Shreyas4991:
@franklindyer : I don't understand the CI additions for update-from-labels. We already have a task for doing this.

franklindyer:
> #158 is in now, so if one can refactor this PR to use that API, that would be great.

Should be good to go now.

franklindyer:
awaiting-review

Shreyas4991:
Some of the proofs are not completed (missing cases). Thus the PR is not successfully building.

franklindyer:
> Some of the proofs are not completed (missing cases). Thus the PR is not successfully building.

It looks like the failure is due to `Completeness.lean`. When I try to build `Completeness.lean` straight from the branch `teorth/main`, it currently fails for me. So I think that file was already broken on the main branch.

Am I missing something?

Shreyas4991:
While I wait for gitpod to do its setup, I believe @codyroux was the author of that file and already fixed everything. Can you try merging main again. Is this PR using an older version of that file. Maybe merging main into your branch helps?

Shreyas4991:
Just saw #263 which got merged. Maybe that helps. I am surprised because I am quite certain we got all errors out of that file and got a greenlight from the CI

Shreyas4991:
Now the main branch needs to be merged into this PR and CI needs to pass before this one is merged.

franklindyer:
> Now the main branch needs to be merged into this PR and CI needs to pass before this one is merged.

I've made a little bit of progress on the issue - it seems like adding `import equational_theories.Homomorphisms` to `FreeMagma.lean` causes `Completeness.lean` to break.

As for why, it's still beyond me. Somehow it makes the `trivial` tactic fail for a couple of proofs where it was previously successful.

codyroux:
It happens, usually a simp gets smarter upstream and makes `trivial`
redundant. I'll fix it but it'll have to wait tomorrow.

Shreyas4991:
It could be fixed in this PR. Only the `trivial` tactic needs to be removed.

Shreyas4991:
@codyroux : I fixed the proofs. It was not an issue with simp. Basically you were calling cases on the hypotheses but we needed them as is.

PR #157, opened 2024-10-01
Add code used to generate the refutations file

carlini:
The FinitePoly refutations have already been added to Lean, but they came from a text file I generated without adding the source that reproduces it. This corrects that, and adds the (relatively simple) script used to generate those polynomials. The script could probably be improved considerably. (I likely will do that at some point.) But for completeness I want to start out by adding the code exactly it was used on the 27th for reproducibility.


vlad902:
Not to give you more work, but I already actually included a copy of refutations.txt in the repo since I needed it. You could git mv it from equational_theories/Generated/TrivialBruteforce/src/refutations.txt and update the path in equational_theories/Generated/TrivialBruteforce/src/generate_refutations.py to `../../FinitePoly/src/equations.txt`

teorth:
Thanks! You could also consider adding the code to the list of scripts at README.md.

PR #158, opened 2024-10-01
homomorphisms and isomorphisms

madvorak:


madvorak:
Out of the last four definitions, we don't want all of them.
Please tell me which one or two we should keep.

madvorak:
I perhaps shouldn't call them "isomorphisms" when the equalities are preserved only in one direction...

madvorak:
propose PR #114

madvorak:
awaiting-review

teorth:
I guess we should follow whatever Mathlib does for the analogous [Mul] and [Add] homomorphisms with regards to making equivalences.

By the way, if you could also update the top-level `equational_theories.lean` to include your new file, and also add a description of your file to the core lean files listed in CONTRIBUTING.md, that would be great.

teorth:
You could add a poll to the zulip poll thread to see if there are any strong feelings on this issue

madvorak:
I will add more API tomorrow.

teorth:
Actually with the likely refactor in which Magma may become closer to Mul this PR may become obsolete.  Stay tuned...

teorth:
There wasn't enough consensus to move from Magma to Mul, so I think we will go ahead with your PR after all.  It looks like you selected a `MagmaHom.toMagmaEquiv`, so all we need to do now is fix the conflict with CONTRIBUTING.md and we are good to merge!

madvorak:
All right, the PR is ready for review now.

Since I don't want to block the issues that depend on this one for longer than need to be, I'll add the remaining API in a separate PR.

PR #165, opened 2024-10-01
Magama Op

codyroux:
Define the "op magma" structure, prove the duality theorem (theorem 3.5).

Uses the `⊧` notation in the meantime, waiting for PR #135 to be merged.


pitmonticone:
The error `MagmaLaw is missing` was due to the fact that you added the `namespace Law`, so now the correct lean declaration to provide is `\lean{Law.Magma}`. 

Fixed.

pitmonticone:
Thanks @codyroux! Merged.

PR #172, opened 2024-10-01
Lean proofs for 4×4 magma brute force

nomeata:
this makes the results from #150 available to lean.


carlini:
(Do not merge yet. As discussed on zulip, this PR is blocking on me to generate the 2x2 and 3x3 magmas to potentially minimize the number of 4x4s necessary. I will handle this tomorrow.)

carlini:
#187 is now pending, and we should discuss how to handle duplicates before merging that one.

PR #177, opened 2024-10-01
Free associative commutative magma model

### LINKED ISSUE #131: METATHEOREM: Formalize and prove Theorem 3.8, giving a necessary condition for implication
This would likely go in a new Lean file (for metatheorems?).  The theorem asserts that if a law has the property that each variable occurs the same number of times on each side, then it can only imply other laws with the same property.  The proof involves testing the law on a characteristic zero field, such as the reals with the addition law.  

It requires setting up some API for counting how often a variable occurs on the LHS and RHS of a law.  It will also require #132 to be completed.

When done, update the blueprint of Theorem 3.8 by adding a \leanok to the statement and proof of the theorem.

codyroux:
We build the model of free associative commutative magmas over a set `A`, which happens to just be `Multiset A`. This should allow us to prove lemma 3.8.

Closes #131


teorth:
Nice!  Can you update the blueprint with the \leanok and the slightly modified proof?  (I agree that using the free associative commutative magma is more natural than using the reals.)

codyroux:
Serendipidity! I just did, didn't rebase though, will fix in the morning if there are issues.

codyroux:
And again for the record, @fpvandoorn devised and proved the crucial lemma.

PR #179, opened 2024-10-01
Invariant based metatheorem blueprint

0art0:
This chapter provides an overview of metatheorems for disproving certain kinds of implications based on invariants of the syntax.


teorth:
Can you expand the list example more precisely - what is the space of lists, what is the magma operation, and what are the lifting maps?

HernanIbarraMejia:
> Can you expand the list example more precisely - what is the space of lists, what is the magma operation, and what are the lifting maps?

I think Anand has gone to sleep (I should too) but here is what he probably means: 

For α a type, let G_α be the set (class?) of all (ordered) lists whose elements have type α. The operation is concatenation. Let ι_α : α → G_α be defined by sending x of type α to the singleton list [x]. Then if f : α → G_α is any function then what the lift of f (which is a function sending lists to lists) does is replace every element of the input list with its image under f and then 'flattens' the output, so we do not end up with a list of lists.

teorth:
I'll merge it for now to add it to the blueprint, but I think some expansion of that particular example can still be added later.

PR #187, opened 2024-10-02
Add all 2x2 and 3x3 magmas via brute force

carlini:
This PR adds all all 3x3 and 2x2 magmas.

I did not deduplicate between magma sizes, for the following reason. Here are the number of equations that are covered if we assume we already have table NxN and then later introduce MxM.

When going from ../data/refutations2x2.txt to ../data/refutations3x3.txt
The number of solved equtions goes from 12567055 to 13602457 for a delta of 1035402
And ../data/refutations3x3.txt has 9 / 299 already covered magmas

When going from ../data/refutations2x2.txt to ../data/refutations4x4.txt
The number of solved equtions goes from 12567055 to 13732566 for a delta of 1165511
And ../data/refutations4x4.txt has 7 / 515 already covered magmas

When going from ../data/refutations3x3.txt to ../data/refutations2x2.txt
The number of solved equtions goes from 13345053 to 13602457 for a delta of 257404
And ../data/refutations2x2.txt has 8 / 10 already covered magmas

When going from ../data/refutations3x3.txt to ../data/refutations4x4.txt
The number of solved equtions goes from 13345053 to 13760346 for a delta of 415293
And ../data/refutations4x4.txt has 100 / 515 already covered magmas

When going from ../data/refutations4x4.txt to ../data/refutations2x2.txt
The number of solved equtions goes from 13732566 to 13732566 for a delta of 0
And ../data/refutations2x2.txt has 10 / 10 already covered magmas

When going from ../data/refutations4x4.txt to ../data/refutations3x3.txt
The number of solved equtions goes from 13732566 to 13760346 for a delta of 27780
And ../data/refutations3x3.txt has 291 / 299 already covered magmas

What this means is that if we included the 2x2 and 3x3 magmas first, we would remove at most 109 (= 100 + 9) of the 515 magmas from the 4x4 table. But if we instead included the 4x4 magmas first, we could cut out *all* of the 2x2 cases, and *291/299* of the 3x3 cases.

Interestingly, but perhaps not surprisingly, there are 28k cases covered by the 3x3 magmas that are not covered by the 4x4. And going from 3x3 to 4x4 covers an additional 4 million. This gives me hope we could find a lot more by going from 4x4 to 5x5, if we found a way to efficiently enumerate the space. But that's a problem for another day.


carlini:
In writing this PR I've identified what might be a bug in the brute force algorithm that appears to make it skip over some refutations incorrectly. While I investigate please do not merge.

nomeata:
I'd still include it, if only to prevent people from thinking “maybe I should try this 3×3 magma”. And 2×2 magmas are very few (6 up to isomorphisms and duals, right), so even if we include them for no additional results, it’s nice to have them for completeness.

Or leave out the 2×2 ones, I have them in `SmallMagmas.lean` anyways.

carlini:
Okay false alarm on any problems. The data was right, my counting was wrong.

To double check, I've re-written a Python version to enumerate all 3x3 tables and confirmed that I get the exact same number of refutations there. This script is much easier to reason about and I'm confident this is correct now.

teorth:
Is this PR still active, or has it been made redundant with the later developments?  If the edit conflicts can be resolved and this PR is not redundant then I'm happy to merge it.

carlini:
I just checked, and adding the 3x3 counterexamples doesn't contribute any new ones after some of the more recent refutations that have been added. 

But this PR only updates the code, doesn't add the Lean theorems. It should still be merged because the code here is cleaned up and I will be using it soon in another PR for hopefully larger counterexample searches.

PR #194, opened 2024-10-02
Add 'raw' subcommand to extract_implications to provide output for other scripts

amirlb:
Ideally we should do all out processing in Lean - single codebase, optimizing compiler.

But for now IMO we should have the ability to run the existing scripts.

### DIFF
--- a/scripts/extract_implications.lean
+++ b/scripts/extract_implications.lean
@@ -6,31 +6,33 @@ import equational_theories.Closure
 
 open Lean Core Elab Cli
 
-def generateUnknowns (inp : Cli.Parsed) : IO UInt32 := do
-  let some modules := inp.variableArgsAs? ModuleName |
-    inp.printHelp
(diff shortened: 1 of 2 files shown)


scripts/extract_implications.lean
```diff
@@ -6,31 +6,33 @@ import equational_theories.Closure
...
-  let some modules := inp.variableArgsAs? ModuleName |
-    inp.printHelp
+def withExtractedResults (imp : Cli.Parsed) (action : Array Entry → IO UInt32) : IO UInt32 := do
+  let mut some modules := imp.variableArgsAs? ModuleName |
+    imp.printHelp
     return 1
   if modules.isEmpty then
-    inp.printHelp
-    return 2
+    modules := #[(ParseableType.parse? "equational_theories").get!]
```
Command-Master, line 14: Does this work?
```suggestion
    modules := #[`equational_theories]
```

PR #195, opened 2024-10-02
INFINITE: State and prove Theorem 2.37 (Equation 5105 has no non-trivial finite models) in Lean

### LINKED ISSUE #181: INFINITE: State and prove Theorem 2.37 (Equation 5105 has no non-trivial finite models) in Lean
The result can be found at https://teorth.github.io/equational_theories/blueprint/subgraph-eq.html#5105-nontrivial

The proof would be similar to the first part of https://teorth.github.io/equational_theories/blueprint/subgraph-eq.html#kis-thm , and go into InfModel.lean.

When done, update the blueprint accordingly with a \leanok for both the statement and proof.

ChienYungChi:
null

Closes #181


ChienYungChi:
awaiting-review

PR #197, opened 2024-10-02
extact_implications: add --equivalence_creators flag to unknowns subcommand

### LINKED ISSUE #188: FRONTEND: Generate a list of "equivalence-creating" implications: open implications, quotiented by equivalence, whose converse is proven or conjectured true
As discussed in https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/Dynamically.20detecting.20useful.20implications.20to.20decide , these are particularly "high-yield" implications to prove true, as each such proof reduces the number of equivalence classes of laws by one, thus potentially eliminating thousands of laws from the list of outstanding implications.  So this would be a good reduced set of equations to test more advanced automated tools against.

If there is already a regularly scheduled task to autogenerate files of implications, this additional file could be added to that scheduled task.

amirlb:
null

Closes #188


amirlb:
awaiting-review

teorth:
Thanks, that was quick!  Could you also add a link to the current output (and comments on the size) in the "Some statistics and data files" section of README.md, for ease of use by other contributors?

amirlb:
Sure, I'll add next time I'm at my computer

teorth:
Also this tool should be integrated somehow with the new dashboard at https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/proposal.3A.20add.20a.20dashboard.20page.20to.20the.20website

amirlb:
> Also this tool should be integrated somehow with the new dashboard at https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/proposal.3A.20add.20a.20dashboard.20page.20to.20the.20website

It's relatively easy to do, at least the counts.
Once we figure out the infra to also save data files to the site it'll be easy to add these there too.

Issue #125, opened 2024-09-30
INFINITE: State and prove Theorem 2.36 (second Kisielewicz law) of the blueprint in Lean.


teorth:
The theorem is at https://teorth.github.io/equational_theories/blueprint/subgraph-eq.html#kis-thm2 .  The proof should be similar to that of https://teorth.github.io/equational_theories/blueprint/subgraph-eq.html#kis-thm , and can be placed in the same file [`InfModel.lean`](https://github.com/teorth/equational_theories/blob/main/equational_theories/InfModel.lean).

When done, a \leanok can be added to the statement and proof of the theorem in the blueprint.

teorth:
Oops, you are right.  Since you are working on the task anyway, please include the relevant corrections to the blueprint and lean files in your PR!

Also, to claim, you need to make a comment containing only the word "claim", as per https://github.com/teorth/equational_theories/blob/main/CONTRIBUTING.md .  (If the instructions there were confusing, let us know and we will try to make them clearer.)

jscanvic:
claim

jscanvic:
propose PR 202

jscanvic:
withdraw PR #202

jscanvic:
propose PR #202

PR #202, opened 2024-10-02
State and prove Theorem 2.36

jscanvic:
* Add theorems `Finite.Equation28394_implies_Equation2` and `Equation28394_not_implies_Equation2` closing #125.
* Introduce Equation28394, a fixed version of Equation28393.


Shreyas4991:
@jscanvic : since this is a draft PR, GitHub won't automatically show this, but an upstream PR might have introduced merge conflicts due to the notation change for Magma. Could you check if merging the main branch into your branch works?

jscanvic:
@Shreyas4991 It appears to be working after changing the notation.

Shreyas4991:
The PR is marked as a draft. The CI will automatically run when the PR is not a draft anymore.

jscanvic:
awaiting-review

jscanvic:
The two equations are indeed completely different. As far as I can tell, giving the new one a new number would simply require changing it in the blueprint `infinite_models.tex`.

Shreyas4991:
Then this would be the path of least resistance. I would recommend it.

jscanvic:
Alright! I've made the changes from `28393` to `28394`.

teorth:
Actually the equation `x = (((y ◇ y) ◇ y) ◇ x) ◇ (y ◇ z)` is numbered 28770, and `x = (((x ◇ x) ◇ x) ◇ y) ◇ (x ◇ z)` is equation 28381.  The full list can be found at https://github.com/teorth/equational_theories/blob/main/data/eq_size5.txt .  Sorry for screwing up the numbering previously, but this should be the canonical numbering to use going forward.

jscanvic:
My bad. I wasn't aware of the extended numbering. It should be good now!

teorth:
Small blueprint issue - one of the references to 28393 in the blueprint now has to be renumbered.  Sorry, that should be the last thing and then we can merge!

jscanvic:
Done!

PR #205, opened 2024-10-02
move dashboard.md -> dashboard/index.md. add links in readme and topbar

dwrensha:
* Moves `dashboard.md` to `dashboard/index.md`. This gives a slightly nicer url (no html suffix) and also provides a natural location to dump images or other assets that we might later want to include in the dashboard.
* Adds a dashboard link in README.md
* Adds a button on the website topbar that goes the the dashboard.

Screenshot from 2024-10-02 14-44-29


dwrensha:
There's currently no way to navigate from the dashboard page to the main page index.html.
I think that should be fixed in a later PR (unless someone has a trivial fix that I could tack on here).

PR #211, opened 2024-10-02
Add explicit implications to Equation 1

vlad902:
Subgraph.lean has a theorem that proves Equation 1 is unconditionally true; however, many tools don't pick that up. Instead of special casing it everywhere, add explicit implications theorems.


teorth:
I think this is fine to have for now.  Once our visualization and implication extraction software stabilizes, we may have a more elegant approach than this, but this is a quick fix that doesn't really create that much bloat in the grand scheme of things.

PR #212, opened 2024-10-02
graphviz tool: always perform a closure->reduction

vlad902:
Generating some graphs, I realized the graphs look far better if they are run with the output of not just a transitive reduction, but a reduction of a closure so that all paths are shortest. It does slow the tool down, but it's still workable.


pitmonticone:
Thanks!

PR #214, opened 2024-10-02
update generate_dashboard to account for conjectures

dwrensha:
Updates generate_dashboard.py to read in the conjecture counts if they are present.

Adds a new table in the dashboard to show these counts.

Screenshot from 2024-10-02 19-33-15


teorth:
Seems clearly worded to me!

PR #216, opened 2024-10-03
Add first small proofs by MagmaEgg, a Rust egg-based automated prover, proving ~1000 implications, 2.8M -> 1.1M unknown

lyphyser:
This is a simple but apparently at least somewhat effective Rust theorem prover based on egg e-graph library written for this project.

It proved 5574 of the 24283 implications in the "only_strongest.txt" file (may be out of date, so some (or most?) could no longer be relevant). This pull requests only includes 3277 implications with a Lean proof of up to 100 lines, since the longer ones are very slow to check in Lean. The "magmaegg-large" branch has the large ones as well.

The code was originally based on the magma_search pull request, but has been pretty much completely rewritten.

Currently search just uses the egg library in a basic fashion, except that in case there are extra variables not present in the LHS, it has code to instantiate them with all subexpressions of the original goal.

Exporting the proofs to Lean has turned out to be harder than finding the proofs, but i managed to write a solution that seems good (modulo some issues in egg that require to sometimes turn off explanation optimization since it sometimes triggers stack overflows and assert failures) and directly produces proof terms using let, Eq.refl, Eq.symm, Eq.trans, Magma.op, a congruence lemma for Magma.op and variables and the hypothesis. I define one letter aliases for them to reduce verbosity.

Possible future work:
- Figure out which implications are important to prove and try it on them
- Replace the fork-based code with self-execution so that it works on Windows and is less of na hack
- Fork egg and fix the buggy optimized explanation generation
- Maybe write Lean code directly instead of writing explanation sexps and converting to Lean code in a second run
- Fix the generation of extra variable values so it doesn't take too much time in pathological cases (i.e. goals with 4-6 variables)
- Determine whether it actually has some advantages compared to Vampire and lean-egg
- Stop the search early if the lhs has been found equal to the rhs
- Support searching for multiple goal equations at once
- Write a custom elaborator for Lean to speed up elaboration
- Support having extra rewrite rules, such as other implications that have been found implied by the hypothesis, or simple equalities found by the egraph search itself
- Run it with massive computing resources if deemed useful and someone offers those


lyphyser:
After reverting Vampire, this are the changes from adding the small theorems.
{"explicit_proof_true": 27803,
"implicit_proof_true": 5294943,
"unknown": 2886650,
"implicit_proof_false": 12994633,
"explicit_proof_false": 829607
}

With MagmaEgg small:
{"explicit_proof_true": 30981,
"implicit_proof_true": 7042768,
"unknown": 1135647,
"implicit_proof_false": 12994633,
"explicit_proof_false": 829607
}

teorth:
Nice progress!  Can you update README.md with a description of the progress in the "Some automated progress", and also correct the most recent "Some statistics and data files" with your new numbers (overwriting the incorrect Vampire data)?  Also, if you could start a brief chapter of the blueprint to basically record the comments you just made above, that would be great.

goens:
That's really cool! I wonder if this should also go in the e-graphs chapter, as it is using the same technique and tool? It would also be interesting to understand how/what it's doing differently than the egg tactic?

vlad902:
Very cool. I took a quick look, and noticed that you can reduce the contributed proofs to just 994 while keeping the same number of transitive implications. You can compute this easily by just following steps 1 and 3 from https://github.com/teorth/equational_theories/tree/main/equational_theories/Generated/EquationSearch

vlad902:
Also one more nit: please add a .gitattributes line as per https://github.com/teorth/equational_theories/blob/main/CONTRIBUTING.md#automated-proofs

lyphyser:
Added documentation and gitattributes, currently applying @vlad902 's procedure to minimize the number of theorems submitted; this also may justify adding some of the theorems with larger and slower proofs, thus giving more proven implications.

Also some of the proofs are probably large because the system automatically retries with explanation length optimization disabled if it causes a crash, which it does quite a bit since the egg implementation is buggy, and shorter proofs might be found once that's fixed.

lyphyser:
I added a new version with only 991 theorems that are enough for the transitive closure to be the same. I removed all the source code and moved it to another branch at https://github.com/lyphyser/equational_theories/tree/magmaegg-code since I'm planning to make significant changes to it.

I propose to merge this for now, and later maybe there will be pull requests with more proofs and better source code .

PR #206, opened 2024-10-02
Use Vampire to resolve almost all unknown implications (without proof)

Command-Master:
Use the Vampire ATP to prove/disprove many implications, transitively resolving almost all implications.
More work is needed to produce Lean proofs for the results, which are currently marked `conjecture`. After this, the statistics are:
explicit_proof_true: 23109
implicit_proof_true: 5295922
explicit_conjecture_true: 38949
implicit_conjecture_true: 1223057
unknown: 7999
implicit_conjecture_false: 1540725
explicit_conjecture_false: 79635
implicit_proof_false: 12994633
explicit_proof_false: 829607

PR #219, opened 2024-10-03
Revert #206

Command-Master:
Revert #206, as a bug caused it to generate false conjectures


teorth:
You need to remove Vampire.tex from content.tex I think

PR #230, opened 2024-10-03
Use Vampire to resolve almost all unknown implications (without proof)

Command-Master:
Now consistently, as far as I can tell, but more scrutiny would be great.

- [x] depends on: #216

Following it:
explicit_proof_true: 30988
implicit_proof_true: 7047455
explicit_conjecture_true: 33159
implicit_conjecture_true: 1066677
unknown: 18034
implicit_conjecture_false: 8312
explicit_conjecture_false: 4771
implicit_proof_false: 12994633
explicit_proof_false: 829607


teorth:
One "unit test" (that may also stretch the range of your conjectures) is to apply duality to your implications (see https://github.com/teorth/equational_theories/blob/main/scripts/find_dual.py or https://github.com/teorth/equational_theories/blob/main/data/dual_equations.md), combine the implications with the duals (there should be extensive overlap), and recompute the transitive closure (I expect a modest increase in coverage, but not much - if you get a dramatic increase, something has gone wrong).

teorth:
How does this PR relate to #281 ?  Does it contain additional conjectural implications not yet formalized in Lean?

Command-Master:
#281 should prove all implications from here, only the non-implications here remain relevant

PR #242, opened 2024-10-03
check axioms of equational results

dwrensha:
Adds a validation step to make sure that theorems tagged with `@[equational_result]` do not
depend on any axioms other than `propext`, `Classical.choice`,  or `Quot.sound`. In particular,
this will report an error on usages of `sorry`.

This adds a little bit of work for each theorem tagged with `@[equational_result]`, but the aggregate
effect is rather small. I observed a clean build to take 8m8s before this change and 8m9s after this change.

### DIFF
--- a/equational_theories/EquationalResult.lean
+++ b/equational_theories/EquationalResult.lean
@@ -1,5 +1,6 @@
 import Lean.Elab.Exception
 import Lean.Elab.Declaration
+import Lean.Util.CollectAxioms
 
 import equational_theories.ParseImplications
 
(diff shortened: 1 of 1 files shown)


dwrensha:
Axioms are discussed here: https://docs.lean-lang.org/theorem_proving_in_lean4/axioms_and_computation.html

> To summarize, then, on top of the underlying framework of universes, dependent function types, and inductive types, the standard library adds three additional components:
    * the axiom of propositional extensionality
    * a quotient construction, which implies function extensionality
    * a choice principle, which produces data from an existential proposition.

PR #244, opened 2024-10-03
Show implication which is only falsified on infinite model

Command-Master:
The implication `Equation3994 -> Equation3588` (`x ∘ y = (z ∘ (x ∘ y)) ∘ z -> x ∘ y = z ∘ ((x ∘ y) ∘ z)`) is only falsified on an infinite model. It's interesting to compare this to `Equation29 -> Equation14` (`x = (y ∘ x) ∘ y -> x = y ∘ (x ∘ y)`) which is true on all magmas and appeared as a Putnam problem.

This was found while working on #230, as there are many implications which Vampire disproved but couldn't produce a finite model for. I suspect that almost all of them are like this, and are only falsified on an infinite model.


Command-Master:
I guess I should also add these proofs to the blueprint

teorth:
Very nice!  Would you be able to add a human readable proof to the blueprint (I guess currently at the bottom of Section 2 is the location for the infinite model stuff, though there are enough results here that it makes sense to spin it off into a separate chapter), and add \lean and \leanok's to link the blueprint proof to the Lean proof?

PR #240, opened 2024-10-03
feat: scoped instances and notation priority

fpvandoorn:
* Adds scoped instances between `Add`, `Mul` and `Magma`.
* Make the priority of `∘` high for `Magma.op`, making it unusable as notation for `Function.comp`.

PR #249, opened 2024-10-03
Switch from \circ to \diw (◇) for magma op

dwrensha:
This unbreaks the current build.

I observe a fresh build to take 1 minutes 20 seconds after this change, compared to 8 minutes previously (before #240).
The reasons are discussed in this thread:
https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/Symbol.20for.20.60Magma.2Eop.60/near/474069724


dwrensha:
I prepared this pull request by doing this:
```
git grep -rl "∘" | xargs sed -i 's/∘/◆/g'
```
and then manually fixing up the results (some instances of \circ needed to remain as such).

Issue #265, opened 2024-10-04
Transitive closure pruning the finite magmas


zaklogician:
In `equation_theories/Generated/All4x4Tables`, we refute potential implications between equational laws by finding small finite counterexamples. These counterexamples are generated by automated tools.

Originally, the generated counterexamples were pruned using the graph of known implications (#196).

The recent addition of 5x5 cancellative magmas broke this pruning process (#224): the file format changed substantially, and the existing pruning script (`All4x4Tables/src/generate_lean.py`) cannot handle it.

Making the pruning work again could substantially speed up the checking of `All4x4Tables`, and further cut down on the number of magmas we need. This could be done by fixing the existing pruning script, or writing a new one (but see the note below regarding the former).

**Note:**

I made an attempt to naively port the script to make it compatible with the changed file format [here](https://github.com/zaklogician/equational_theories/commit/63355c23bbcbb0d9ac16951ebf27ff2dca6a6c0a). At first everything seemed to work, the proof goes through. However, inspecting the output of `lake exe extract_implications outcomes --hist` shows that we lose quite a few refutations! Unfortunately, there are many places where things could go wrong, including in the original script, one of my changes, or even in the Lean code which counts implicit proofs. As of now I have no spare capacity to look into this myself, but am happy to assist if somebody takes it up.

teorth:
Is this task now completed, or should the issue remain open?

zaklogician:
This was resolved by #275 .

PR #271, opened 2024-10-04
Fix bug in counter example pruning script

nomeata:
Strangely, even after updating `implications.json` to the latest version, the unknown count increases from from 2726304 to 2734065. Not sure what’s going on, but as this is definitely fixing one (among more than one?) bug, probably still worth merging.

This does not update the All4x4Magmas files, because I’m confused how they should be generated these days, see
https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/4.C3.974.20magmas/near/474767268


zaklogician:
The `Refutation*.lean` files checked into the current main branch are generated using the [generate_5x5_refutations.sh](https://github.com/teorth/equational_theories/blob/finite_magma_tools/equational_theories/Generated/All4x4Tables/src/generate_5x5_refutations.sh) bash script from the file [plan-5x5.txt](https://github.com/teorth/equational_theories/blob/finite_magma_tools/equational_theories/Generated/All4x4Tables/data/plan-5x5.txt) (itself generated from other files by a heap of tools starting with Mace4, although a copy is checked in so we don't have to rerun it every time).

The bash script itself calls out to a Haskell tool, `gen-refutations` in `src/finite_magma_tools` which parses `data/plan-5x5.txt` and emits the Lean code directly.

However, `gen-refutations` does not read `implications.json` and does not do pruning - it's a throwaway tool, meant to eventually be replaced by an updated `generate_lean.py` which can handle something like `plan-5x5.txt`. I made an attempt at doing precisely this (see the repo linked in #265) but bounced back when the implication counts didn't match up.

I think this bug might have been the cause, so I'll revisit #265 by rebase my port over this once this is merged.

PR #272, opened 2024-10-04
Inlcude MagmaEgg implications in build (and fix-up application symbol)

vlad902:
The previous PR adding these implications (#216) did not include the implications in Generated.lean so they haven't gotten picked up by the main build and all the subsequent tooling (dashboards, etc.)

Add it to Generated.lean, and fix-up the changed application symbol.


lyphyser:
Oops, sorry, I think I lost that change when rebasing.

PR #275, opened 2024-10-04
All4x4Tables: transitive closure pruning for 5x5 magmas (#265)

zaklogician:
I implemented `make-plan` output format support in `generate_lean.py`. This means that it can now be used to prune the refutation set of the cancellative 5x5 magmas using the already proven implications read from `implications.json`.

This greatly reduces the number of explicit proofs we have to do:  of the previous 196990 facts that we had to check explicitly, pruning leaves only 24266 (a 87% improvement).

Since #271 fixed the bug that led to lost implications, this now put us back on 2726304 unknowns.

Merging this will resolve issue #265 .


equational_theories/Generated/All4x4Tables/src/generate_lean.py
```diff
@@ -146,22 +159,47 @@ def «{name}» : Magma (Fin {div}) where
...
+            if line.startswith("Magma") or line.startswith("Satisfies") or line.startswith("Refutes"):
+                row.append(line)
+            if len(row) == 3:
+                all_rows.append(row)
+                row = []
+    return all_rows
+
+rows = []
+with open(f"{dir}/data/plan.txt") as f:
+    rows = create_rows(f)
```
nomeata, line 192: No need to use `with` if its only used once anyways:
```suggestion
rows = create_rows(open(f"{dir}/data/plan.txt"))
```

equational_theories/Generated/All4x4Tables/src/generate_lean.py
```diff
@@ -146,22 +159,47 @@ def «{name}» : Magma (Fin {div}) where
...
+with open(f"{dir.parent}/All4x4Tables.lean", "w") as main:
```
nomeata, line 193: We should eventually rename this folder (and merge with the other finite examples), but not in this PR :-)

equational_theories/Generated/All4x4Tables/src/generate_lean.py
```diff
@@ -146,22 +159,47 @@ def «{name}» : Magma (Fin {div}) where
...
+  for i, row in enumerate(rows):
+      leanfile = f"{dir}/Refutation{i}.lean"
+      data = parse_row(row)
+      if data and data["div"] < 10:
```
nomeata, line 197: Do we have larger magmas in the data? If not,

```suggestion
      if data:
```

equational_theories/Generated/All4x4Tables/src/generate_lean.py
```diff
@@ -146,22 +159,47 @@ def «{name}» : Magma (Fin {div}) where
...
+        data = prune_row(data)
+        print(f"Writing {leanfile}")
+        main.write(f"import equational_theories.Generated.All4x4Tables.Refutation{i}\n")
+        with open(leanfile, "w") as f:
+              f.write(generate_lean(data))
```
nomeata, line 202: I guess I thought I’d be passing the handle to `generate_lean` and then in the end didn’t
```suggestion
        open(leanfile, "w").write(generate_lean(data))
```

equational_theories/Generated/All4x4Tables/src/generate_lean.py
```diff
@@ -146,22 +159,47 @@ def «{name}» : Magma (Fin {div}) where
...
-          leanfile = f"{dir}/Refutation{i}.lean"
-          data = parse_row(line)
-          if data and data["div"] < 6:
-            data = prune_row(data)
-            print(f"Writing {leanfile}")
-            main.write(f"import equational_theories.Generated.All4x4Tables.Refutation{i}\n")
-            with open(leanfile, "w") as f:
-                  f.write(generate_lean(data))
-
+def create_rows(f):
```
nomeata, line 162: The logic is now spread (even more) in `create_row` and `parse_row`. Worth cleaning up, or are we going to make all tools use the same output format anyways?

nomeata (commented):
Minor comments, fine to just merge if you prefer

equational_theories/Generated/All4x4Tables/src/generate_lean.py
zaklogician: In the long run, the upcoming json format will replace all of the ad-hoc formats developed here.
But this is a good point, I'll have a quick think and see if I can push all the case analysis into create_row.

zaklogician: Well, I tried, but I was not really satisfied with the outcome. The rewrite ended up being 6 lines longer, and imho harder to read.
I say let's kick this can down the road :)

equational_theories/Generated/All4x4Tables/src/generate_lean.py
zaklogician: Agreed.

PR #276, opened 2024-10-04
Finite counterexamples generated by the Vampire theorem prover

Command-Master:
Solves ~8,000 unknown implications


zaklogician:
Very nice!

Note: we had an accidental regression in `All4x4Tables` after #271 (my review wasn't careful enough, so a different, older set of refutations ended up being pushed). So if you ran this in the last 3 hours, some of the anti-implications might already have counterexamples in `All4x4Tables` which will be restored by #275.

Command-Master:
> Very nice!
> 
> Note: we had an accidental regression in `All4x4Tables` after #271 (my review wasn't careful enough, so a different, older set of refutations ended up being pushed). So if you ran this in the last 3 hours, some of the anti-implications might already have counterexamples in `All4x4Tables` which will be restored by #275.

I ran it before that, so it should be fine

zaklogician:
Btw: having taken a look, it looks like Vampire managed to generate a large number of different tables. Is each example constructed independently, or are there some checks to see whether the resulting tables refute any implications apart from the one in the problem statement?

If the former, you can create an issue for this after the merge: I'll eventually get around to recovering the generated tables and putting them into `All4x4Magmas` so that we'll know all the implications they refute.

nomeata:
> Any idea what might be causing the `(kernel) deterministic timeout`? It didn't happen on my computer

Proving that a `Fin 10` magma satisfies some equation with a few variables may be too slow to do for the `decideFin!` tactic.

Maybe try `by native_decide` instead for large magmas and small number of facts (only 2 here).

This may need some more engineering to make tractable.

Command-Master:
I've noticed that MemoFinOp is using a base which isn't a power of 2, I could try to change that (and make it use bitshifts and bitwise and) and see if that helps

nomeata:
I have strong doubts that this is the problem. It's still two operations, so the same deterministic timeout would apply. In fact, with shift you need an extra operation to test/clamp to `Fin n`.

I think it’s just the sheer number of operations.

Profiling first would help.

equational_theories/MemoFinOp.lean
```diff
@@ -9,39 +9,41 @@ namespace MemeFinOp
...
-def opOfTable {n : Nat} (t : Nat) (a : Fin n) (b : Fin n) : Fin n :=
-  let i := a.val * n + b.val
-  let r := (t / n^i) % n
-  ⟨r, Nat.mod_lt _ (Fin.pos a)⟩
+def opOfTable {n : Nat} (s t : Nat) (a : Fin n) (b : Fin n) : Fin n :=
+  ⟨(t >>> (s * (a.val * n + b.val))) % n, Nat.mod_lt _ (Fin.pos a)⟩
 
 /-- `enum n` is the array of all elements of `Fin n` in order -/
-private def Fin.enum (n) : Array (Fin n) := Array.ofFn id
+private def Fin.enum (n) : Array (Fin n) := Array.ofFn Fin.rev
```
nomeata, line 16: You probably should update the docstring as well (or remove it for this private helper)

Command-Master: I undid this change, indeed it didn't help

.gitattributes
```diff
@@ -10,4 +10,4 @@ equational_theories/Generated/FinSearch/theorems/*.lean linguist-generated
 equational_theories/Generated/FinSearch.lean linguist-generated
 equational_theories/Generated/MagmaEgg/*/*.lean linguist-generated
 equational_theories/Generated/MagmaEgg/*.lean linguist-generated
-equational_theories/Generated/VampireProven/Proofs*.lean linguist-generated
+equational_theories/Generated/VampireProven/*.lean linguist-generated
```
nomeata, line 13:
```suggestion
equational_theories/Generated/VampireProven.lean linguist-generated
equational_theories/Generated/VampireProven/*.lean linguist-generated
```

scripts/transitive_reduction_nonimplications.rb
```diff
@@ -0,0 +1,23 @@
...
+
+# Given as an input the JSON produced by `lake exe extract_implications --json`, outputs
+# a transitive reduction of the non-implications
+
+require File.join(__dir__, 'graph')
+
+graph = Graph.new
+data = JSON.parse(File.read(ARGV[0]))
+data["implications"].each { |s|
+  graph.add_edge([false, s["rhs"][8,10].to_i], [false, s["lhs"][8,10].to_i])
```
nomeata, line 11: Is this `[8,10]` extracting the number from the definition name? But what if the number is longer than two characters? Or is this something else?

Command-Master: It takes a substring of length 10 starting from the eighth character

nomeata: Ah, it's length, not end. TIL :-)

zaklogician:
I can confirm that the minimization has resolved the OOM issues I experienced earlier, and I am now able to build tihs on my laptop.

PR #280, opened 2024-10-04
extract_implications: sort equations in outcomes output

amirlb:


equational_theories/Closure.lean
```diff
@@ -138,35 +138,51 @@ def Bitset.set (b : Bitset) (n : Nat) : Bitset :=
...
-      for lhs in satisfied ++ refuted do
-        match eqs.containsThenInsertIfNew lhs eqs.size with
-      | (n, neqs) =>
-        eqs := neqs
-        unless n do eqs_order := eqs_order.push lhs
-    | _ => pure ()
-  pure (eqs, eqs_order)
+      for eq in satisfied ++ refuted do
+        eqs := eqs.insert eq
+    | .unconditional eq =>
```
dwrensha, line 173: Nice. After this change, `Equation1` starts showing up in the extracted outcomes for `Subgraph`, as it should.

teorth:
I guess sorting can't hurt, although it seems that for some applications it is not necessary.

dwrensha:
:+1: I like this change. It cleans up a few things, and it makes the output of `extract_implications outcomes` a bit easier to use.

PR #281, opened 2024-10-04
Proofs generated using the Vampire theorem prover

Command-Master:
Reduces the number of unknown implications for 1.1M to 29688. Also add a term elaborator for superposition calculus, which is used for the proofs.


vlad902:
Cool! Based on looking at a couple of files, I think this set of proofs maybe reducible? Also one nit: add a .gitattributes as per CONTRIBUTING.md

vlad902:
I actually pulled your change down to check if any of my local implications that I haven't submitted yet are still necessary, and I found that is reducible. Apologies if I've made a dumb mistake. 

```
# On main *without* your change
lake build
lake exe extract_implications --json equational_theories.Generated equational_theories.Subgraph > main_implications.json
cat main_implications.json | ruby -rjson -e 'JSON.parse($stdin.read)["implications"].each { |s| puts s["lhs"][8,10] + "," + s["rhs"][8,10] }' | sort -u > main_implications.csv

cat Generated/VampireProven/*.lean | ruby -ne 'puts "#{$1},#{$2}" if $_ =~ /Equation(\d+)_implies_Equation(\d+)/' | sort -u > new_theorems.csv
cat main_implications.csv new_theorems.csv | sort -u > new_state.csv
ruby scripts/transitive_reduction.rb new_state.csv | sort -u > new_state.reduced.csv
comm -13 main_implications.csv new_state.reduced.csv > new_theorems_to_include.csv
wc -l new_theorems_to_include.csv new_theorems.csv
```

shows 1775 vs 5768

Command-Master:
Ah, I only did the reduction using the theorems this proves, not using other existing theorems :facepalm: . I'll rerun it doing the reduction properly

teorth:
Is this ready to merge?  If all the proofs compile in Lean then we are pretty safe, the question is whether there is any further savings in compile time to be made by further reduction.

teorth:
I guess if there are size issues we can always refactor later, so I'll push the button now.  Looking forward to seeing the new state of the graph!

teorth:
I just realized that there is no README.md announcement of the Vampire results that advanced the completion percentage so substantially.  Could you make an addition to that file accordingly?  It would also be good to describe the methodology of the Vampire run in the blueprint.

Command-Master:
> I just realized that there is no README.md announcement of the Vampire results that advanced the completion percentage so substantially. Could you make an addition to that file accordingly? It would also be good to describe the methodology of the Vampire run in the blueprint.

I made #302

Issue #147, opened 2024-09-30
TACTIC: law_to_equation and equation_to_law


teorth:
Write two tactics, tentatively called `law_to_equation` and `equation_to_law` (but the names can be changed, that do something like the following).

Functionality of `law_to_equation`:

1. If the goal is currently of the form `LawX \geq LawY` for two laws `LawX`, `LawY`, change the goal to `∀ (G: Type *) (hG Magma G), EquationX G → EquationY G`, where `EquationX` and `EquationY` are equational counterparts of `LawX`, `LawY`.  (There may be some universe issues regarding the type of `G`.  Also one has to decide whether the conversion of `LawX` to `EquationX` is via changing the name (and assuming that `LawX`, `EquationX` are already defined), or by some syntactic conversion procedure.)
2. If the goal is currently of the form `¬LawX \geq LawY` for two laws `LawX`, `LawY`, change the goal to `∃ (G: Type *) (hG: Magma G), EquationX G ∧ ¬ EquationY G`.

Functionality of `equation_to_law` is to invert the previous tactic.  Namely:

1. If the goal is currently of the form `∀ (G: Type *) (hG Magma G), EquationX G → EquationY G`, change the goal to `LawX \geq LawY`.
2. If the goal is currently of the form `∃ (G: Type *) (hG: Magma G), EquationX G ∧ ¬ EquationY G`, change the goal to `¬LawX \geq LawY`.

Optionally, allow these tactics to also apply to hypotheses as well as to goals.

franklindyer:
Note: this would be enormously helpful for duality-based proofs. Since dualizing a law is implemented syntactically, proofs-by-dual (like my example POC proof-by-dual in #122) involve a lot of converting back-and-forth between equations and laws. These tactics should make this kind of proof much less clunky.

Command-Master:
claim

Command-Master:
This requires refactoring the logic part to work with `Type*` instead of `Type`

teorth:
This is what #186 was intended to fix.  There is the alternative of refactoring the entire rest of the project to work with `Type` instead of `Type*`, but my hope was that the equivalence between the two would allow one to avoid this.

Command-Master:
I'm not quite sure how this solves it—currently you can't even write `G ⊨ law` is `G` isn't in the first universe.

teorth:
[...]

PR #285, opened 2024-10-04
Metatheorem about duals

goens:
Adds some "metatheorems" about op and symm, in particular `implies_op` which is the "duality" theorem that lets us collect additional "implied" implications.

Will add these implied implications also to the count in a separate PR (as that depends on #147)


teorth:
I'll merge this once the `MagmaLaw.lean` conflicts are resolved.

goens:
should be resolved now @teorth

PR #291, opened 2024-10-04
Add Z3Counterexamples.lean, with a counterexample found by z3

dwrensha:
Resolves a new edge that is still not covered by #281, plus 5 more by transitivity.

Includes in a comment the z3 spec that I ran, which involved bit vectors of length 3.
When I tried the same thing with bit vectors of length 2, z3 said it was unsatisfiable.

### DIFF
--- a/equational_theories.lean
+++ b/equational_theories.lean
@@ -12,3 +12,4 @@ import equational_theories.OrderMetatheorem
 import equational_theories.Preorder
 import equational_theories.SmallMagmas
 import equational_theories.Homomorphisms
+import equational_theories.Z3Counterexamples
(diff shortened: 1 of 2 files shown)


teorth:
Is it worth adding this counterexample to an existing lists of finite magmas?

dwrensha:
The list I am aware of is in SmallMagmas.lean, and this magma did not seem appropriate for there ("This file defines some trivially small magmas...")

Your comment did prompt me to learn about `calculate_facts`, which does indeed find some more edges where this magma helps, but in a highly verbose form that needs reduction. Since I don't have time to make that nice now, I just added a TODO here about trying `calculate_facts` on this magma.

PR #293, opened 2024-10-05
fix(equational_theories/Preorder.lean): partial revert of #273 "reverse pre-order direction"

agjftucker:
…rder direction"


teorth:
I think it has stabilized in the right orientation now... but certainly it is a good idea to do these sorts of "unit tests"!

Issue #278, opened 2024-10-04
EXPERIMENTAL: Explore the finite magma refutation data and see how the number of variables in the law play a role


teorth:
Discussed in https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/visualization.20of.20graph.20closure .  Let k denote the number of variables in a law.  As discussed above, the k=1 laws should be underdetermined and obeyed by a large number of finite magmas; k=2 laws should be determined and obeyed by a few number of finite magmas; and k>2 laws should be overdetermined and obeyed by mostly just the trivial magma.  Now that we have data sets of what laws are obeyed by which finite magmas of order up to 4, plus some selected larger magmas, some data analysis should be able to see whether this heuristic holds up, and whether there are anomalous laws that exhibit a different number of solutions than the heuristic predicts.

When done, report on the outcome of the experiment in a new chapter of the blueprint.

teorth:
One can also explore the statistics of the implication graph stratified by k.  For instance, the above heuristic predicts that the k=1 laws should have many laws implying them, but few consequences, whereas the k>2 laws should have the opposite.  It should in particular be quite rare for a low k law to imply a high k law, though this is not an absolute prediction (e.g. one can artificially introduce more variables via substitutions).

Issue #266, opened 2024-10-04
CENTRAL_GROUPOID: Resolve all remaining implications of the form 168 => X


teorth:
Only 42 implications of the form Equation 168 => Equation X remain open.  Equation 168 is the equation of a central groupoid.  The discussion in Section 5 of Knuth's paper https://www.sciencedirect.com/science/article/pii/S0021980070800321?ref=cra_js_challenge&fr=RR-1 appears to resolve many of these, as does his examples of "non-natural central groupoids", such as those in Section 7 of Knuth, which could be added to the current lists of interesting magmas, as discussed in https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/Collecting.20interesting.20finite.20magmas.

zaklogician:
claim

zaklogician:
Progress report. Knuth's A_2 resolves a whole bunch of these, but not quite all of them. The following 13 remain open:
3456, 3457,3511,3512,3513,3862,3877,3918,3955,3993,5105,28393,374794.

Next step: try to construct the other noncentral magmas;  if that fails for some reason, try some hand-guided generation to produce counterexamples.

teorth:
I guess this means that A_2 also creates some additional anti-implications beyond 168=>X then?  But perhaps those were already all known.

zaklogician:
Yes, it resolves a few non-168 implications as well. The reason there is no PR yet is that I plan to put `A_2` into SmallMagmas instead of `All4x4Tables` (where it'd just be one more multiplication table without context).

teorth:
With the latest update, it seems from https://teorth.github.io/equational_theories/implications/?168 that we have now in fact resolved all implications (from what I can tell, all the open implications were in fact true!).

teorth:
One can of course continue formalizing results about central groupoids, but it seems that this particular issue is in fact now resolved.

zaklogician:
Sweet, thanks!

Issue #180, opened 2024-10-01
DATA: Create a text file to store interesting finite magmas of size 5 or greater - such as (Z/6Z,+), and automated tools to use this text file to generate anti-implications.


teorth:
More specifically, create a script to test the magmas in the text file against the list of outstanding implications, and add any anti-implications generated to the list of generated equations.  Also, create CI so that if additions are made to the text file, then the script is run again.

Discussion at https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/Collecting.20interesting.20finite.20magmas

carlini:
Note that (Z/6Z, +) and other similar equations should already be included in the set of finite polynomial refutations generated by brute force because it searches over equations of the form `ax^2 + by^2 + cx + dy + exy`. (This was not completely exhaustive search, but it was pretty close, and I've just checked that nothing new would be implied by it that wasn't already covered by the current polynomials.)

teorth:
Onh ok. Then Z/6Z can just be used as a test case to check that it doesnt rule out any still open implications, but hopefully people will contribute more interesting finite magmas soon.

carlini:
In case someone's curious, I believe the following two implications from the finite_poly generations "cover" `(x + y) % 6`:
`(3 * x**2 + 3 * y**2 + 1 * x + 1 * y + 0 * x * y) % 6` 
and
`(0 * x**2 + 2 * y**2 + 0 * x + 2 * y + 3 * x * y) % 4`.

teorth:
The discussion at https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/Refutations.20with.20finite.20cancellative.20magmas is also relevant.  We should have a standardized system for accepting lists of finite magmas (either small human-generated files or large computer-generated files) in a standard location, so that automated tools can test them for refutations, and update CONTRIBUTING.md to describe the procedures and standards for doing so.

teorth:
See the discussion in #294 for a description of the current options to incorporate finite magmas into the automated refutation generator.  The main thing currently missing is a user-friendly interface to input them.

teorth:
Closed as finite magma explorer already contains these instructions.

PR #294, opened 2024-10-05
All4x4Tables: update after #281

zaklogician:
This PR:

**1.** Adds 4 more cancellative magmas that were found through hand-guided search and which refute 31 of the 29688 implications still open after #281, bringing down the number to 29657. The idea was to find for cancellative magmas that specifically fail to satisfy large one-variable laws. The addition of the new magmas also results in a slightly better plan (with fewer Lean checks required).

**2.** Re-runs the transitive closure pruning using the large number of recently proved implications. This further pushes down the number of explicitly proven implications from 604701 to 581162 .

**3.** Adds a minor QoL improvement to variable counting in `finite_magma_tools`, which makes it more useful for #278 -style experimentation.


teorth:
Is there a standard location where one can input other medium size magmas for  automated testing, such as the 8-element magma in #291 or the 9-element magma being used in #266?

zaklogician:
@teorth

> Is there a standard location where one can input other medium size magmas for automated testing, such as the 8-element magma in #291 or the 9-element magma being used in #266?

I'll probably get around to building something more user-friendly for this in the near future, but the current situation is this:

`All4x4Tables` has the `data` directory where we normally put these tables and where the proof planner can read them from. [[Here]](https://github.com/teorth/equational_theories/blob/main/equational_theories/Generated/All4x4Tables/data/cancellative-manual.txt) is an example of such a file, but as you can see it already has to contain all the equations satisfied in the given magma. (FinSearch has something similar - the two dirs will eventually be merged, but we'll have to sort out how this goes).

The `parse-mace4` tool in [ finite_magma_tools`](https://github.com/teorth/equational_theories/tree/main/equational_theories/Generated/All4x4Tables/src/finite_magma_tools) can create this list in the required format from Mace4's outputs. This is not that hard to hand-create from a multiplication table. Here's an example, whitespace does not matter:

```
interpretation( 6, [number=1, seconds=0], [
        function(+(_,_), [
			   1, 0, 3, 2, 5, 4,
			   2, 3, 5, 4, 1, 0,
			   4, 5, 1, 0, 2, 3,
			   0, 1, 2, 3, 4, 5,
			   5, 4, 0, 1, 3, 2,
			   3, 2, 4, 5, 0, 1 ])
]).
```

teorth:
Ultimately one could imagine a web interface where one could upload (in some standard format) a multiplication table, the tool displays all the equations it satisfies, checks if it creates any new refutations, and if so it accepts the table and adds it to some standard directory from which the refutations are generated.  There may be a few thousand finite magma refutations left out there, so this might be worthwhile.  Anyway, I made a link here from #180 which one can re-interpret as asking for this sort of user-friendly functionality.

PR #295, opened 2024-10-05
Refactor equations to use EquationsCommand in Equations.lean

### LINKED ISSUE #290: REFACTOR: Make the equation definitions in Equation.lean use the `equation` elaborator from EquationsCommand.lean
This will help standardize the introduction of equations, in particular making it easier to transfer equations from AllEquations.lean to Equations.lean.  (Also, in the future we will want to pair each equation with a law, so having it all done through the standardized `equation` elaborator will mean we do not need to do this by hand for each subgraph equation.

Shreyas4991:
null

Closes #290

### DIFF
--- a/equational_theories/Equations.lean
+++ b/equational_theories/Equations.lean
@@ -1,5 +1,5 @@
 import equational_theories.Magma
-
+import equational_theories.EquationsCommand
 /-! List of equational laws being studied -/
 
 /-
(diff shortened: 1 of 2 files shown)


Shreyas4991:
Docstrings need to be enabled on the `equation` command. For some reason this is a syntax error in lean. More details on [the comment on the issue](https://github.com/teorth/equational_theories/issues/290#issuecomment-2394852354)

PR #299, opened 2024-10-05
bug fix: don't write "🎉 All steps completed successfully!" after a failing `run_before_push.sh` command

euprunin:
Fixes the following issue:

```
% bash ./scripts/run_before_push.sh
…
❌ Error: […]
Press any key to exit...
./scripts/run_before_push.sh: line 70: return: can only `return' from a function or sourced script
🎉 All steps completed successfully! You are ready to push.
```


pitmonticone:
`exit 1` isn't good in this case because, given the default VS Code setup many users are adopting, it would close the terminal.

pitmonticone:
Thank you very much for raising the issue concerning the final message. I'll fix as soon as I find the time.

pitmonticone:
@euprunin can you specify which error actually allowed that to happen. 

It should just stop, ask you to press any key and then close.

pitmonticone:
Thanks @euprunin, merged.

PR #301, opened 2024-10-05
A simple web UI to view equations and their implications

carlini:
This PR adds the web UI for the equations and their implications. The code is pretty awful, 90%+ of it was written by Claude/ChatGPT but it works.

The main thing I'm unsure about is the CI component of this. I've never tried to do that before, and while I think it works, I'm not actually sure. So if someone who understands that would be able to take a look at that before merging that'd be great.

It also does not link from the home page yet. I'll add that link once we confirm that everything is working properly.


teorth:
Assuming there are no CI issues I would say this is pretty ready to merge.  Could you update README.md to change the link to the old explorer tool to the new one (and remove most of the text around that link)?

carlini:
I was going to play it safe and leave the README as-is just until I can verify that the page works with the CI build, and then submit a second PR linking to it from the project page and from the README. (But if you'd rather just give it a shot all at once I can make the changes now.)

teorth:
OK, let's try merging it and seeing if it works, then you can send a second PR to link it up to the documentation.

carlini:
Looks like it's all working. Will update the documentation now.

PR #305, opened 2024-10-05
Display Hasse diagram for the Subgraph

### LINKED ISSUE #183: FRONTEND: Use CI to auto-generate a Hasse diagram of the equations in Equations.lean, and display this image in README.md and blueprint
Ideally, the image should be stored in some static location (presumably in the Images folder), and README.md and Figure 3.1 of the blueprint can point to that location (and the old manually-constructed Hasse diagram can be dropped from the blueprint as being obsolete).  NOTE: the Hasse diagram should be oriented according to the updated guidance at CONTRIBUTING.md

vlad902:
Also update the script to generate such graphs.

Closes #183


vlad902:
I forgot to update the blueprint, will do that now.

pitmonticone:
Thanks a lot, @vlad902! 

`leanblueprint pdf` returns an error regarding the image we're including.

Please consider running the executable script `scripts/run_before_push.sh` to avoid this risk.

vlad902:
Ah, of course. It builds successfully locally, because I have the PNG, but it fails in CI, because I forgot to git add it.

Issue #143, opened 2024-09-30
DATA: Add a MagmaLaw associated to each equation in Equations.lean or AllEquations.lean


teorth:
For each EquationX `abbrev` in Equations.lean, also add a `LawX` that defines the corresponding element of `MagmaLaw`. (EDIT: this step will no longer be necessary if #290 is completed.)

Modify the `equation` elaborator in EquationCommand.lean to do something similar for AllEquations.lean.

This is the first step towards transitioning from having equations as the ground truth of the universe of assertions, to laws.

goens:
claim

goens:
propose PR #357

0art0:
Apologies for the overlap, my PR #356 also does something similar, as a path towards proving non-implications using meta-theorems. I'd be happy to close my PR and make the code depend on your PR.

PR #322, opened 2024-10-05
`law_to_equation` and `equation_to_law`

Command-Master:


goens:
@Command-Master this is really nice! Can I ask if you'd mind splitting the functionality from the actual commands? That way I could use the `equation_to_law` functionality in the elaboration of the `equation` command to also generate laws, for #143?

PR #324, opened 2024-10-05
All4x4Tables: Integrate straggler counterexamples

zaklogician:
Expands the number of finite magmas in All4x4Tables by integrating a couple of finite magmas which were previously used in counterexamples elsewhere.

**Benefits:**

* For all these examples, each satisfied and refuted equation was determined,
      which settled 236 previously unknown anti-implications.
* These examples were put through the proof planner, meaning that while the
      number of magmas serving as counterexamples increased greatly (from 543 to 740) the number of checks the
      kernel has to perform remains about the same. The `generate_lean.py` script can
     also prune them using transitive closure whenever new implications are proven.
    
**The newly added sources:**

     * Magmas constructed by the Vampire theorem prover gathered from #276.
     * Two magmas constructed by the Z3 solver gathered from #308.
     * A 5x5 magma manually constructed by Pace Nielsen.

 With the importation of the straggler examples, the current set of Vampire disproofs become redundant, so these are removed. The commit changes the VampireProven documentation with a bit of explanation about how Vampire `Disproof*.lean` files will appear when Vampire disproofs are generated. It also updates the blueprint with more sources of counterexamples.

Unknowns at 21679.


Command-Master:
How did you produce the lists of which implications each table proves? I'm running Vampire with more time per equality, so I want to add the magmas it finds directly to here

PR #325, opened 2024-10-05
Make equations link to a 'prove this' link.

### LINKED ISSUE #316: FRONTEND: A web tool that, given equation numbers X and Y, creates a Lean Web playground for proving that X does or does not imply Y
This is best illustrated by an example.  If one input the equation numbers "43" and "4512" into the tool, the tool should then direct the browser to [this link](https://live.lean-lang.org/#codez=MYGwhgzhAECyYHMC2ZoApCNwNAXNAKgJ4AOAptAPoCU0A7gBYkBOJAUNNAPZE7RaBJhL2gCMLFgEsAdgDMxAD2wA2AKzQARNEDjpGugBeAHxxEKAHRdRYAEYXmAN2gBRAI4BXMABcxHCQBYAzOgBxXEJSCmoAbXhkVACAXRwdaEAAImhZaAIeAIAaVM103XytWXMrWwcXd08fJQBGACZA4OIyKmhIoxj47ESUtIyAL0yctK00DK1+6kS0EfTqCdE3Bg5mJHLXDy8/cjEkIhAxEghyJw2q71qGtCD8ZrC2qJRoOPQ6XFPKrf8A6neKzeq9WeCWgFgyEBWjAIi2Wq3Wnx8vnIEg4bh2ewORxO/3Olx4gGAiRq3UKtNDkXCPGKUHIfAF+YGAciJoAAa+F0vEBEFgtjQCGMKEsIA), which provides a minimal import necessary to state the assertions that Equation43 does, or does not, imply 4512.  If a user proves one of the assertions (hopefully not both), this could be cut and pasted into a future PR if the implication is new.

This tool would ideally be folded into #209: each implication in #209 could be supplied with a link to the playground generated as above. However, creating a standalone tool as proof of concept may be an initial step of this task.

carlini:
The web UI now has links next to every unknown proof that you can click that will drop you into a lean prover.

Fixes #316

### DIFF
--- a/home_page/implications/index.html
+++ b/home_page/implications/index.html
@@ -196,6 +196,7 @@
 
     <script src="implications.js?6"></script>
     <script src="find_equation_id.js"></script>
+    <script src="to_lean.js"></script>
     <script src="script.js"></script>
   </body>
 </html>
--- a/home_page/implications/script.js
+++ b/home_page/implications/script.js
@@ -307,7 +307,9 @@ function renderImplications(index) {
 
 	[forwardStatus, backwardStatus].forEach((status, statusIndex) => {
             const isConjectured = status.includes('conjecture');
-            const item = `<div uid=${i} class="implication-item ${isspecial} ${status} ${isConjectured ? 'conjectured' : ''}">${eq}${more_same}</div>`;
+	    let proofhref = gen_proof_url(index, i);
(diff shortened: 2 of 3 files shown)


teorth:
Wow that was a lot quicker than I expected!

PR #327, opened 2024-10-05
add outcomes png to dashboard

dwrensha:
Adds the output of `outcomes_to_images.py` to the dashboard page.

Makes `generate_dashboard.py` responsible for launching subprocesses for whatever intermediate artifacts it needs. This way, the dashboard generating in `blueprint.yml` is reduced to a single line that calls `generate_dashboard.py`.

Screenshot from 2024-10-05 14-15-24

### DIFF
--- a/.github/workflows/blueprint.yml
+++ b/.github/workflows/blueprint.yml
@@ -133,8 +133,7 @@ jobs:
 
       - name: Generate home_page/dashboard.md
         run: |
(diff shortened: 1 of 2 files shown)


teorth:
These are intriguing visual artefacts in the graph.  Some periodicity and some extra structure near the diagonal.  Not sure what to make of all that!

PR #330, opened 2024-10-05
All4x4Tables: 2531 does not impy 1075

zaklogician:
I give a 13x13 counterexample showing that neither of the dual pair 1076/2531 imply 1075. The constructed magma ends up resolving over 1% of the remaining unknowns.

This dual pair had by far the highest unknown count among those equations that are not implied by anything else (except for x=y).


teorth:
Fascinating!  Was the 13x13 example constructed by hand, or created by Z3, or some mix of both?

PR #357, opened 2024-10-06
Equations elab laws

### LINKED ISSUE #143: DATA: Add a MagmaLaw associated to each equation in Equations.lean or AllEquations.lean

goens:
This adds elaboration of `MagmaLaw`s to the `equation` command, so we can have both with one command.

Closes #143 

Depends on #355


teorth:
I saw the discussion at https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/Automatically.20generating.20magma.20laws.20.26.20proofs.20from.20equations - does that cause any issues with merging this PR?  Or can we use this PR as a starting point to help coordinate the other equation <-> law efforts that are going on right now?

goens:
I think the consensus with @0art0 was that we'll merge this first, and then recover the two directions using #368

PR #373, opened 2024-10-06
Add conjectures produced by Vampire

Command-Master:
These conjectures are created from implications which are proven to be false by saturation, which we don't know how to formalize in Lean yet.
Seeing as `Equation3964_not_implies_Equation3456` from #362 was originally in here, I don't think an off-by-one situation like occurred previously is likely (these are also transitively consistent with all existing results).


teorth:
One can also check to see if these conjectures are approximately symmetric with respect to duality as an additional "unit test".

PR #381, opened 2024-10-06
add --extra flag to extract_implications

dwrensha:
Adds logic to ignore equations outside of the core set of 4694. The filtering is by default enabled in `extract_implications`. The old behavior can still be accessed via a new `--extra` flag.

This will make the dashboard display information only about the core set.


teorth:
One could perhaps add a second page to the dashboard folder to hold the "extra" statistics, and have a link from the "main" dashboard to the "extra" dashboard.  But this is low priority.

PR #386, opened 2024-10-06
Fix second half of proof of Theorem 3.4 in infinite_models.tex

avi-levy:
This PR is a continuation of #347 which was fixing issues in the blueprint presentation of Theorem 3.4. In this PR, I move on to fixing a logic error in the argument that the presented magma is indeed a model of the law being considered. The logic error lies in the claim that `(1 \op x) \op w = x` whenever `w` is a power of two. This claim does not hold for the magma under consideration, since `(1 \op 1) \op 2 = 4` rather than `1` as claimed. Fortunately a weaker claim suffices. While working on this, I observed that the choice of magma is not unique and if one wished to optimize for proof elegance, there may be other magmas that admit simpler proofs. However, I did not pursue this because the benefit appears to be less than the cost of introducing clashes with the other places this specific magma is used (such as in the Lean implementation and in the literature).

### DIFF
--- a/blueprint/src/chapter/infinite_models.tex
+++ b/blueprint/src/chapter/infinite_models.tex
@@ -46,14 +46,21 @@ we conclude that $f_y$ is also a constant function.  But this function is alread
 
 To construct an infinite model, consider the magma of positive integers $\Z^+$ with the operation $x \op y$ defined by
 $$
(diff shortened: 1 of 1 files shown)


teorth:
Thanks!  (This is a good example of why we formalize our proofs :).)

PR #387, opened 2024-10-06
Fix_links

teorth:
Make links in README.md global for benefit of web page; also fix up some blueprint tags


pitmonticone:
If you want to make them global for the webpage, then you need to change the home_page/index.md file instead.

pitmonticone:
I'm currently keeping the index.md updated copy-pasting the README.md, but I know it's not at all ideal. 

I'll write a script to automate the copy-pasting of the relevant text in the README.md to the index.md.

PR #389, opened 2024-10-07
Invariant-based counter-example files

0art0:
These are automatically generated counter-example files based on invariants of the syntax (following chapter 8 of the blueprint).

It should soon be possible to also generate proofs for these claims automatically.


teorth:
These are quite large files, and may significantly increase the total compilation time if they are added to the Lean codebase.  Is it possible to perform transitive reduction on them?  Or extract only those implications that are not already established as a separate file, and only import the latter into the codebase (the former can be in a non-Lean format, such as a JSON).

Also, for the purpose of replication and documentation, it would be good if you could supply the code you used to generate these files, as well as a description of exactly what metatheorems you are using here, and how you are applying them to generate the files.

PR #390, opened 2024-10-07
proof of eq3588 not implies eq3994

dwrensha:
This is the same as the proof of "eq3994 not implies eq3588", but with x and y swapped.
As far as I know we don't yet have an easy way to get a dual proof like this, without this kind of code duplication.


dwrensha:
See #391

PR #396, opened 2024-10-07
better description for getting a random open problem

madvorak:


madvorak:
Yes. Should it stay there?

madvorak:
We might as well say:
"Why not try proving a random open problem?"

pitmonticone:
I think your "Why not try proving a random unknown implication?" is good.

Thanks! Merged.

PR #401, opened 2024-10-07
Add Graphiti: a tool to view graphs

vlad902:
A preview of this tool can be tested here: https://tsyrklevi.ch/graphiti/

This tools adds graphiti to the home_page, but does not yet link it from the README.md or the dashboard. I would like to first ensure that the data is generated correctly and the webapp loads as expected, and only then introduce links.


pitmonticone:
Thanks! Merged.

PR #402, opened 2024-10-07
LinearOps: Solve positive facts using `ring_nf` (26s → 10s)

nomeata:
and use `decideFin!` otherwise.

Discussion: https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/An.20old.20new.20idea/near/475253132


pitmonticone:
Thanks @nomeata! Merged.

PR #405, opened 2024-10-07
move 953 implies 2 as it was not found by lean-egg

goens:
moving it just to keep track of things properly (not to unfairly claim this nice proof for the automated tactic)


pitmonticone:
Thanks!

PR #423, opened 2024-10-08
Add commentary below equations in equation explorer

carlini:
Adds any commentary from commentary files in a commentary/ directory to the equation explorer.

Screenshot 2024-10-07 at 22 13 19


teorth:
Unfortunately this feature is glitchy.  Firstly, when one moves from an equation with a commentary to an equation without a commentary, the old comment stays around.  Furthermore, when one then moves back to an equation with a commentary... something strange happens; the main equation name doesn't change, but the commentary does.

With the current commentary, there is now some redundancy as the comment files also mention the equation number and shape, but actually it's worth keeping this around for debugging purposes for the moment.  I can go and manually edit the commentary files later.

PR #438, opened 2024-10-08
qa: run lean4lean check as part of `run_before_push.sh` and `pre-push.sh`

euprunin:


pitmonticone:
Can you please discuss your `lean4lean_check.sh` file [here](https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/CI.20Lean.20Environment.20Check). 
I know @Shreyas4991 is working on it.

Shreyas4991:
One suggestion. I already have a python script in works in #421. If you make this PR depend on that, then you can simply use that script

Shreyas4991 (commented):
Please make this PR dependent on #421 and use that python script.

Shreyas4991:
Another important point: Please follow the claims process before making PRs.

Shreyas4991:
Git python is only used in the virtual env. One problem with the bash script is that it pins the commit of lean4lean. In our repository we run an auto update action to keep us on the latest mathlib toolchain. Thus inevitably the lean4lean toolchain will become incompatible. This is why I made the choice to stay on the main branch of lean4lean.

Shreyas4991:
Another problem with this script is that it builds all of lean4lean including the theoretical developments and verifications a la metacoq. This is not only not necessary, but it can also cease to build because of updates of the toolchain of lean4lean. Thus my script builds only the CLI tool.

Shreyas4991:
> Pinning a version helps maintain consistency and reliability by avoiding unexpected updates, ensuring that version bumps are intentional and introduce changes carefully.
> 
> I suggest we wait for third-party code review feedback from a software engineer before moving forward. As mentioned, I have no preference whether my code is used or not, as long as the solution is robust and adheres to software engineering best practices.

Pinning a version is fine in usual software development, but in lean projects, we have found it much more important to stay consistent with the latest mathlib toolchain, since mathlib evolves rapidly. This repository is set to auto update its toolchain, and this happens very frequently. It is extremely important for the project lean-toolchain and lean4lean's lean-toolchain to match.

Shreyas4991:
> > Another problem with this script is that it builds all of lean4lean including the theoretical developments and verifications a la metacoq. This is not only not necessary, but it can also cease to build because of updates of the toolchain of lean4lean. Thus my script builds only the CLI tool.
> 
> Are you sure about that? This script runs `lake -q build lean4lean`. Is that building more than we actually need?


Yes, I am sure. I discussed this extensively with Mario on zulip. Also I just noticed that this bash script doesn't limit the number of threads. This is super critical for lean4lean.

Shreyas4991:
More broadly "software engineering best-practices" are heuristics that don't always apply in all situations identically.

Shreyas4991:
> > > > Another problem with this script is that it builds all of lean4lean including the theoretical developments and verifications a la metacoq. This is not only not necessary, but it can also cease to build because of updates of the toolchain of lean4lean. Thus my script builds only the CLI tool.
> > > 
> > > 
> > > Are you sure about that? This script runs `lake -q build lean4lean`. Is that building more than we actually need?
> > 
> > Yes, I am sure. I discussed this extensively with Mario(the creator of lean4lean) on zulip. I also ran a number of tests on his locally before adding my PR.
> 
> I reviewed the Python script:
> 
> ```
> buildret = subprocess.run([lakePath.as_posix(), "build", "lean4lean"]).returncode
> ```
> 
> Isn't that equivalent to `lake -q build lean4lean`?
> 
> Could you clarify which specific build command you’re suggesting?

Right, I am on my mobile but it seems your script is also building only the lean4lean cli. Sorry about that. 

Nevertheless, pinning versions is definitely not going to work well. I prefer a CI that tells us that lean4lean cannot run and passes over a CI action that is definitely going to run for more than a few days until the next toolchain update arrives in the project. Also lean4lean is very memory expensive unless it is only allowed to use a few threads.

Shreyas4991:
So, Mario gave the advice independent of my full dependencies run. Lean4lean fills up memory rather quickly if it is allowed to spawn threads. Also this project has a number of large files and that is likely to grow.

Shreyas4991:
Thanks. For the CI I am most likely going to stick with my script with any improvements you might suggest. 

 However my PR doesn't add any pre push hooks yet, so I would like to include your hooks. If you wish, we can use your scripts for the hook.

PR #440, opened 2024-10-08
Graphiti: Add option to show unknown edges

vlad902:
As per [1], add the ability to graph unknown implications with dotted edges, and make it easy to navigate to this graph from the equation explorer.

[1] https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/Equation.2065.20-.3E.20left.20cancellability/near/475352370

### DIFF
--- a/.github/workflows/blueprint.yml
+++ b/.github/workflows/blueprint.yml
@@ -153,7 +153,8 @@ jobs:
       - name: Generate home_page/graphiti/graph.json
         run: |
           ~/.elan/bin/lake exe extract_implications --json > /tmp/implications.json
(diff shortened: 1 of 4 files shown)


teorth:
Are some of the implication arrows now downward pointing in your example? If so it might be worth adding in the arrows to orient the graph.  Certainly 359 doesn't imply 1491, for instance.

vlad902:
While answering this question and testing it locally, I realized that due to a slight error in how I used graphviz, the resulting graph was more chaotic than necessary. By changing the way I used graphviz, I was able to generate much less chaotic graphs when unknowns are included.

vlad902:
So to answer your original comment, perhaps including arrow directions to the implication edges is now unnecessary. The unknowns can still point up of course, but that seems reasonable.

PR #445, opened 2024-10-08
qa: run `scripts/lean4lean_check.py` as part of `run_before_push.sh` and `pre-push.sh`

euprunin:
This is a Python alternative to #438. It does not introduce any new Python dependencies.

PR #446, opened 2024-10-08
Revert "qa: run `scripts/lean4lean_check.py` as part of `run_before_push.sh` and `pre-push.sh`"

teorth:
Reverts teorth/equational_theories#445

Apologies for not following the procedures we had laid out in CONTRIBUTING.md.  It's probably best to start over with a fresh PR and have it go through the review process.


teorth:
To clarify: I approved the PR thinking that it created a new script rather than modifying existing scripts, but that was based on my incorrect understanding of the PR.  Sorry again for the trouble caused.

PR #447, opened 2024-10-08
qa: add `lean4lean_check.py` and run as part of `run_before_push.sh` and `pre-push.sh`

euprunin:


Shreyas4991:
This PR is identical to #445 . Is there a reason this is being repeated?

PR #457, opened 2024-10-09
consistent `equation` spacing

madvorak:
Most declarations that use the keyword `equation` have two spaces on each side of `:=` so I put it everywhere consistently.


pitmonticone:
I think it's better to adopt the usual convention of one spaces everywhere.

madvorak:
I like people started putting double spaces around `:=` as it makes it more readable.

PR #459, opened 2024-10-09
Lift implications laws

goens:
Lift implications on Equations to Laws in Meta code.

This is part of #174

PR #460, opened 2024-10-09
Split examples of FactSyntax, remove unnecessary dependency in FreeMagma

goens:
Makes `FactSyntax` not depend on `AllEquations` and `FreeMagma` not depend on `EquationalResults`. We need this to avoid cyclic dependencies on meta code. This is a cleanup that I'm factoring out of #459


goens:
oops, CI failed, should've checked that - sorry. I'll fix it

PR #466, opened 2024-10-09
[DO NOT MERGE] counterexample found by Nikita Kalinin

madvorak:

### DIFF
--- a/equational_theories/Nameme.lean
+++ b/equational_theories/Nameme.lean
@@ -1,0 +1,14 @@
+import equational_theories.AllEquations
+import equational_theories.EquationalResult
+import Mathlib.Tactic.FinCases
+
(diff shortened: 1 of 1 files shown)


madvorak:
@npkalinin

madvorak:
TODO which file should it be moved to?

madvorak:
do not merge

madvorak:
It seems that @npkalinin disproved an implication that had already been disproved because the webpage gave us a template for the opposite direction than what is claimed to be an open problem.

Can somebody confirm please?
Bug report: https://github.com/teorth/equational_theories/issues/469

madvorak:
Somebody please check after me that the non-implication is already known, and close this PR afterwards.

PR #471, opened 2024-10-09
Typo fix in Definition 7.1 in asterix.tex

avi-levy:
Corrected from "The function $f$ is a bijection from $E_1 \backslash E_0$ to $E_2 \backslash E_0$." to "The function $f$ is a bijection from $E_1 \backslash E_0$ to $E_2 \backslash E_1$."


avi-levy:
awaiting-review

PR #427, opened 2024-10-08
Equations command modification

0art0:
This PR modifies the `equation` command to automatically generate the corresponding law with variables over `Fin n` (in addition to the law with variables over `Nat` that is already being generated). 

This is used to show that the automatically generated laws are equivalent to the original equation in terms of satisfiability. 

Additionally, this PR introduces an environment extension to store all the equational laws, which can be used to export laws as described in #142. 

This PR also modifies a result transporting laws from `Fin n` to `Nat` to omit the `n != 0` hypothesis.

Issue #142, opened 2024-09-30
DATA: Create an automated tool to regularly gather the equations from Equations.lean and AllEquations.lean and output them to a text file that is easy for machines to read


teorth:
This would presumably already be a byproduct of #29, but is worth isolating separately.  The model here is that `Equations.lean` and `AllEquations.lean` represent the ground truth of the universe of equations/laws under consideration, so that new equations can be added to the universe by updating the lean files accordingly

Command-Master:
Currently #126 only considers equations that appear in some implications, so this will require some change

teorth:
It would also be nice to have a separate file that only lists the implications and anti implications on the equations defined just in Equations.lean

vlad902:
Recently, Equations 28393, 374794, and 5105 have appeared in the output of `lake exe extract_implications` breaking some of my tools as they expected all equations to be in AllEquations.lean (I parse out the ones that are commented there as well.) That's probably a fault of my tooling, but also a good reason to do this to make it easier. Everyone can load a JSON.

Command-Master:
That also broke some of my tools. I could add a Lean script to extract the equations, what would be a good I/O for it?

vlad902:
If you mean an output format, just a JSON hash seems easy for everyone to consume. Are you thinking of including it in `extract_implications` or somewhere else?

Command-Master:
I think `extract_implications` already does a lot of things, it's better to make another script.
By output format I meant how to describe the equations. Is something like `{"Equation3": [0, [0, 0]], "Equation4": [0, [0, 1]]}` good? Would it be better to have `{"Equation3": "x = x ◇ x", "Equation4": "x = x ◇ y"}`? Something else?

teorth:
As a short term fix one can of course add these equations in commented form to AllEquations.lean, and also add a note sonewhere in CONTRIBUTING.md requesting that one add the commented equation to AllEquations.lean whenever adding a novel equation to Equation.lean as a courtesy.

vlad902:
> I think `extract_implications` already does a lot of things, it's better to make another script. By output format I meant how to describe the equations. Is something like `{"Equation3": [0, [0, 0]], "Equation4": [0, [0, 1]]}` good? Would it be better to have `{"Equation3": "x = x ◇ x", "Equation4": "x = x ◇ y"}`? Something else?

[...]

PR #472, opened 2024-10-09
split AllEquations into pieces for more parallelism

digama0:
While more can be done to improve the performance of the `equation` command, one thing that helps a lot is to make an embarrassingly-parallel task more parallel in practice by splitting it into separate files. The `Equation` files have also been moved into a common folder.


carlini:
This might break some of the scripts that load equation data from this file. Ideally they wouldn't be doing that in the first place and should read from a separate python-friendly file. If things don't build for this reason I can try to to fix things.

digama0:
Can you point me to the code that does this?

carlini:
There are several I'm aware of:

https://github.com/teorth/equational_theories/blob/c27789ec7ea82c2d870080bf8c5bc0ff3af12a51/scripts/generate_graphiti_data.rb#L16

https://github.com/teorth/equational_theories/blob/c27789ec7ea82c2d870080bf8c5bc0ff3af12a51/scripts/explore_magma.py#L13

https://github.com/teorth/equational_theories/blob/c27789ec7ea82c2d870080bf8c5bc0ff3af12a51/scripts/find_dual.py#L169

https://github.com/teorth/equational_theories/blob/c27789ec7ea82c2d870080bf8c5bc0ff3af12a51/scripts/generate_equation_implication_js.py#L109

Shreyas4991:
@digama0 : A lot of the automated diagram generating python and ruby scripts assume the existence of AllEquations.lean because this structure was agreed on near the beginning of the project. They read this file as a textfile. This isn't super ideal and maybe we should have written an export script in lean that exports the equations of "AllEquations.lean" into json. But fixing this now will be non-trivial.

teorth:
This is what #142 is supposed to do.  Perhaps #427 can be adapted to create such a JSON, which all other tools then draw from?  Then if we ever refactor the equation lean files again then one just has to change the one export tool.

carlini:
(I'd be happy to help rewrite the scripts to load from an appropriate json data file as appropriate.)

Shreyas4991:
The [script that generated them] appears deterministic, so in principle, the same script can be modified to output the json file. However the script writers must then use this json file.

vlad902:
I can update the ruby code when I'm back at the computer later today.

digama0:
Note: the current state is that all scripts have been fixed, although I don't know how to use them all so they are only lightly tested.

vlad902:
> Note: the current state is that all scripts have been fixed, although I don't know how to use them all so they are only lightly tested.

Ah, apologies for the confusion.

Issue #171, opened 2024-10-01
FRONTEND: Develop a CI to regularly compute the latest statistics of implications, and add them (e.g., as a time series) to README.md (or some other file linked from README.md).


teorth:
Relevant statistics could include those from #29, as well as number of outstanding implications modulo known equivalences.  Links to text files enumerating these various implications would also be useful.

Shreyas4991:
Should we consider this task completed?

teorth:
I don't think we actually have a time series yet - as far as I know the dashboard output is not being saved in any file -  but we should have all the architecture to do so, and I think it is a good idea (I am manually recording the dashboard status each day now, but this is clearly not the long-term solution).

carlini:
I've started a PR #479 to add this data for all past commits. Currently it has going back to the 2nd, and I'll get the commits from before then later today.

The main question is how to make it automatically add this data for the future. It appears there are three options:

1. Because main is protected, the CI tool can't push anything to main. One option around this is to have it push a commit to the branch that's about to be merged before the merge happens. I think this is possible, but have never done it. So maybe it doesn't work.
2. It might be possible to unprotect main only for the CI tool. I don't know how to do this, but some internet searching suggests that there are ways to grant special permissions that might make this work. A maintainer would have to handle doing this though.
3. A final option is to have the CI tool push to a separate (unprotected) branch of the project every time someone merges a new commit with main, and then have someone manually merge this separate branch into main every once and a while (once a day, say) to keep things on track.

Thoughts?

Shreyas4991:
The script can be called from inside CI?

Shreyas4991:
One problem with option 3 is that the project webpage is generated off the main branch of the project.

carlini:
I think the CI can add additional commits to a PR. So I'd run the tool, store the results, and log it to a file, and commit that to the repo. Or I read somewhere that was possible.

As for option 3: agreed, this would make it only update once a day. But at least it would be automated and would just need someone to click "merge" instead of rerun it.

Shreyas4991:
Can't the CI make a pull request with a specific label?

carlini:
Yeah, I guess that could happen too, but would then double the number of commits to main.

Shreyas4991:
Can we run this as a cron job instead?

PR #479, opened 2024-10-09
[DRAFT] Automatic tracking of statistics after each commit

carlini:
This PR is the first step to log data about how far the project has progressed over time. Right now I've run the `extract_implications --hist` tool for every commit from when the tool was created until this morning. I'll start back-filling the remaining commits by cherry-picking the extraction tool into the past.

Once I've completed the backfill, I'll try to get the CI to automatically add one of these files with each PR. I have an idea for how to do this but it'll take a bit of work. and will discuss this on issue #171.


carlini:
Ah no. It looks the file existed at that point, but didn't support --hist. I'll backport that too.

teorth:
It's strange that the number of open problems increased from 10-02 to 10-03... possibly some flaw in the transitive closure algorithm?

teorth:
@carlini Are you still intending to work on this PR?  If not, we can reassign the task.

PR #490, opened 2024-10-10
Add script to generate TPTP for any implication, extracted from Vampire scripts

lyphyser:
TPTP is the standard format for theorem provers, used for instance by the Vampire script.

This is useful to run other automated theorem provers (or Vampire with custom settings) on the remaining implications.

Might be also be worth it to generate TPTP for all current open implications and provide it either on the repository or on the webpage, perhaps automatically with CI.


vlad902:
It'd be nice if this had a very simple description at the top, even just an example invocation so it was clear how to use it. Also a line in the README.md scripts section so others could find it.

PR #492, opened 2024-10-10
Results on Central Groupoids

### LINKED ISSUE #352: CENTRAL_GROUPOID: Prove Theorem 5.9 (natural central groupoid axiom)
This result of Knuth shows that a certain five-operation law (Equation 26302) is equivalent to being a natural central groupoid.  The proof is quite lengthy; a sketch is at https://teorth.github.io/equational_theories/blueprint/sect0004.html#natural-central-groupoid , but one should review the original paper of Knuth (Theorem 5) at https://www.sciencedirect.com/science/article/pii/S0021980070800321 for the full details.  But perhaps there is also a more efficient proof.

This would most naturally go into `CentralGroupoids.lean`.  When done, mark the theorem and proof of Theorem 5.9 with `\leanok`'s and an appropriate `\lean{}` tag.

b-mehta:
Closes https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/Central.20groupoids. 

This PR contains a number of results on central groupoids and natural central groupoids.

This is a draft PR.

- [ ] Ensure that @zaklogician's work is not destroyed here - probably the simplest way is to move that file to NonnaturalCentralGroupoids.lean
- [ ] Give an example of an implication that reduction can disprove
- [ ] Add a module doc which more clearly highlights the key results here


zaklogician:
This is very nice work. Re point 1: would same file but separate namespace be a viable solution? I think it'd avoid cluttering up the main `equational_theories` directory too much.

b-mehta:
Unfortunately not, the imports I need from mathlib clash with some of your results

PR #496, opened 2024-10-10
DATA: Add examples to smallest_magma.txt

### LINKED ISSUE #475: DATA: Create an enriched version of the  `Smallest Magma` file that lists not only the size of the smallest finite magma obeying a law, but the actual magma itself.
Suggested in #443

The smallest magma file is [here](https://github.com/teorth/equational_theories/blob/main/data/smallest_magma.txt)

few:
This fixes #475 by 

a) Modifying `data/smallest_magma.txt` to include an example of a magma of smallest size
b) Introducing `data/small_magma.txt` for examples of small magmas where it is not known if the magma has the smallest possible size
c) Introducing `scripts/gather_small_magmas.py` that can read various files to extract small magma examples and add them to `data/smallest_magma.txt` or `data/small_magma.txt`

Currently `equational_theories/Generated/All4x4Tables/data/` and `equational_theories/Generated/VampireProven/` are read. The two examples in `data/small_magma.txt` come from the first folder.


teorth:
Would it be possible to still keep the original text file that just has the size of the smallest magma in its current location, and add an additional file that also has the smallest magma itself?  The original text file might still be useful for some applications, and easier to use, even if it contains strictly less information.

few:
I have renamed the files to `data/smallest_magma_examples.txt` and `data/small_magma_examples.txt` and kept the old one. The script updates the old files as well (that's causing the reordering there).

PR #503, opened 2024-10-11
Equation1661 does not imply Equation1657

zaklogician:
This takes care of 28 unknowns altogether. For details about the construction see the [Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/1661.20-.3E.201657.20-.20another.20one.20bites.20the.20dust).

I decided to put the proof in its own file, `InfModel_1661.lean`, since the argument is somewhat long: we have to check by a thorough case analysis that we do not break 1661 via our truncation/deformation. These also require separate lemmas, which should not be kept as `have lemma : ...` inside the theorem environment for performace reasons (most `simp`s would time out). All in all, putting it directly in `InfModel.lean` would have cluttered up that file.

(Remark: This is actually my second take on the proof: my first was an induction argument checking the operator table for larger and larger squares, but it turned out slower and harder to follow, with just as many case analyses, albeit different ones. I mention this just in case anyone else looks at this proof and considers going by that route: it's been explored.)


zaklogician:
Note: the description in the blueprint is deliberately coarse for now. Once the Asterix PR is merged, I'll modify it to link to the relevant section and briefly explain that these magmas are "patched up in finitely many cases".

PR #510, opened 2024-10-11
Introduce rule_system to define a system of confluent rewrite rules and disprove all? implications that can be made confluent

lyphyser:
This greatly cuts down on the boilerplate and allows to feasibly disprove all implications that can be turned into confluent rule systems, including those with more than 2 rules


lyphyser:
It should now have disproofs for all the non-implications modulo duality that are turned into confluent rule systems by current Vampire or Twee.

The missing ones should all be duals to the included ones.

lyphyser:
I split everything into files according to how many rules were in the system, and refactored the single-rule code to replace all the duplicated proofs with custom tactics.

Should be in relatively good shape now.

PR #511, opened 2024-10-11
All unknowns 1661 -> ... resolved

zaklogician:
This is an extension of #503.

Following a suggestion to do the same for the Asterix-Obelix counterexample ([here](https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/Equation.2065.20-.3E.20left.20cancellability/near/476097979)), I considered whether my patch construction provides counterexamples to the other implications from 1661. Long story short, they do, resolving another 46 unknowns.

PR #512, opened 2024-10-11
clean up InfModel_1661.lean a bit

dwrensha:
* Replaces some proofs by `omega`, which works and is very fast on these lemmas.
* Removes some lemmas in favor of already-existing theorems, found by the `exact?` tactic.
* Rewrites some non-terminal `simp` as `simp only`.


zaklogician:
Thanks, this is much appreciated!

See discussion regarding compatibility w/ #511 is [here](https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/1661.20-.3E.201657.20-.20another.20one.20bites.20the.20dust). I think it'll be a few minutes to fix.

dwrensha:
@pitmonticone this is ready to merge now

PR #522, opened 2024-10-11
Fix issue #507

### LINKED ISSUE #507: EQUATION_EXPLORER: Fix glitches in the commentary feature
When navigating to an equation with a commentary, the standard heading of the equation (listing the equation number, form, and dual) is missing.

When navigating from an equation without commentary to an equation with commentary, the standard heading of the old equation is retained even though the display otherwise switches to the new equation.

When navigating from an equation with commentary to an equation without commentary, the commentary from the previous equation is retained.

(Note: I think the correct functionality should be that an equation page should always display the standard header (including the duality information), and also the commentary if it exists.  At present, the commentary also somewhat duplicates the header information, leading to some redundancy, but I plan to clean up the commentary files to remove this redundancy once this glitch is fixed.

vlad902:
This change clears commentary between equation transitions in the equation explorer, and also adds some styling to make the commentary visibly distinctive from other elements.

image

Closes #507

### DIFF
--- a/home_page/implications/index.html
+++ b/home_page/implications/index.html
@@ -28,6 +28,11 @@
       .header .equation-stat { cursor: pointer; }
       .header .equation-stat:hover { background-color: #e0e0e0; }
       .checkbox-container { margin: 10px 0; }
+      #equationCommentary {
+        border: 1px solid black;
+        padding: 5px;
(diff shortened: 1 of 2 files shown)


vlad902:
awaiting-review

PR #524, opened 2024-10-11
prove duals of 1661 anti-implications via duality metatheorem

dwrensha:
This resolves 74 currently "unknown" implications.

### DIFF
--- a/equational_theories/InfModel.lean
+++ b/equational_theories/InfModel.lean
@@ -1,4 +1,6 @@
 import equational_theories.Equations.All
+import equational_theories.MagmaOp
+import Aesop
 import Mathlib.Data.Fintype.Card
 import Mathlib.NumberTheory.Padics.PadicVal.Basic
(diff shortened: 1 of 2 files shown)


pitmonticone (approved):

dwrensha:
I'm realizing now that this does not actually use a duality metatheorem (something like [`implies_iff_op`](https://github.com/teorth/equational_theories/blob/8d331edb7356c97415c3fb31b9910d839cfe40f9/equational_theories/MagmaOp.lean#L61)). Making better use of such a theorem could potential eliminate some of this boilerplate code.

teorth:
If you see a relatively easy way to utilize a duality metatheorem right now, we can wait for that update to approve this request, otherwise we can merge for now (I don't think this will create an undue burden on the compilation time) and wait for a future refactor to replace these dualized proofs by some semi-automated tactic that can dualize an existing proof, or something.

PR #526, opened 2024-10-11
Lifting magma families

### LINKED ISSUE #377: METATHEOREM: Non-implication proofs using invariants
Generating conjectures and proofs of non-implications using invariants generated from lifting magma families, following [Chapter 8 of the blueprint](https://teorth.github.io/equational_theories/blueprint/sect0007.html).

0art0:
This PR formalizes the theory of Chapter 8 and uses it to generate several non-implications with formal proofs.

The relevant files are in `equational_theories/Generated/InvariantMetatheoremNonimplications`.

Closes #377


teorth:
Roughly how many implications are being added here?  The main Lean codebase is being transitively reduced to reduce compilation times, so we would likely only add implications to the main folders if they settle new implications or anti-implications, but I don't see harm in adding a lot of redundant implications so long as they are not placed in the main codebase but in another directory (e.g., in the data directory).

0art0:
The counts for the files are

```
6070 - instLiftingMagmaFamilyLeftProj_counterexamples.lean
30 - instLiftingMagmaFamilyList_counterexamples.lean
160 - instLiftingMagmaFamilyMultiset_counterexamples.lean
6070 - instLiftingMagmaFamilyRightProj_counterexamples.lean
```

While generating these files, I am also filtering to make sure that the results are not already contained in the closure of the established equational results. However, these counts are still much higher than I would expect after filtering, so I can take another look at my code to see if I am generating the closure correctly (if I don't manage to figure out the reason for these high numbers, I'd be happy for these files to go in the `data` directory instead).

0art0:
After pulling in the latest changes from `main`, it seems like the code produces no results that haven't already been established (I've added a bit of code to ensure that if this script is run twice consecutively, the results from the previous run don't block out results from the new run). I would still like to go ahead with merging this PR since the general framework here is likely to be useful in settling other kinds of non-implications.

0art0:
Strangely, running the script yet again (after pulling in the most recent changes) gives the same results as before with the same frequencies (my hunch is that the results from the first run of the script were blocking the ones from the second).

0art0:
awaiting-review

0art0:
Looking more closely at the outputs, they seem to mainly involve equations from `Equations/Basic.lean` about which fewer results have been established in general (presumably). It's likely that none of these results are among the 700 or so remaining open implications/non-implications, which (if I recall correctly) are about the initial set of 4694 equations.

0art0:
Apologies for the series of messages above, here is a summary:

The counts of the results in the generated files are
```
6070 - instLiftingMagmaFamilyLeftProj_counterexamples.lean
30 - instLiftingMagmaFamilyList_counterexamples.lean
160 - instLiftingMagmaFamilyMultiset_counterexamples.lean
6070 - instLiftingMagmaFamilyRightProj_counterexamples.lean
```

While generating these files, I am also filtering to make sure that the results are not already contained in the closure of the established equational results.

The reason these counts seem higher than one would expect from the current status of the project is likely because the script is also considering the sporadic equations defined in `Equations/Basic.lean` which are not among the initial set of 4694 equations.

teorth:
It might still be possible to transitively reduce the ~12000 implications you generated because some of them may imply some of the others in combination with the existing implications, even if none of them are in the transitive closure of the existing implications itself.  For instance, if it was already known that A => B, and you establish that A !=> C and B !=> C, then the second anti-implication can be dropped as it already follows from the first, even if both anti-implications are new.  I think there are already transitive reduction scripts that can be applied to the union of the existing implications and your new ones which should see if there are significant reductions to be made here.

0art0:
@teorth Thank you for the suggestion, I've implemented this kind of transitive reduction and it has reduced the sizes significantly. Here are the updated counts:
```
45 - instLiftingMagmaFamilyLeftProj_counterexamples.lean

10 - instLiftingMagmaFamilyList_counterexamples.lean

60 - instLiftingMagmaFamilyMultiset_counterexamples.lean

55 - instLiftingMagmaFamilyRightProj_counterexamples.lean
```

Issue #501, opened 2024-10-10
lake exe export_equations doesn't build


vlad902:
I get the following build error on `lake exe export_equations` on main:
```
error: ././././scripts/export_equations.lean:27:39: application type mismatch
  Json.arr fun h ↦ Subtype.map (fun x ↦ ?m.281 x) h laws
argument
  fun h ↦ Subtype.map (fun x ↦ ?m.281 x) h laws
has type
  (∀ (a : RBNode Name fun x ↦ Unit), RBNode.WellFormed Name.quickCmp a → ?m.273 ((fun x ↦ ?m.281 x) a)) →
    Subtype ?m.273 : Sort (max 1 ?u.261)
but is expected to have type
  Array Json : Type
error: ././././scripts/export_equations.lean:27:52: invalid constructor ⟨...⟩, expected type must be an inductive type with only one constructor
  RBNode Name fun x ↦ Unit
error: Lean exited with code 1
Some required builds logged failures:
- scripts.export_equations
error: build failed
```

_Originally posted by @vlad902 in https://github.com/teorth/equational_theories/issues/142#issuecomment-2405393119_

vlad902:
And while we're here, is it possible to have this write to stdout instead of a fixed file name? It's not a given that it will be invoked from the root dir of the project repo, and it better matches the output of `export_implications`

amirlb:
claim

amirlb:
I ran out of idea trying to get this script to work. It's simple to fix the types, and to change inputs and outputs. However

* `magmaLawExt.getState env` returns an empty list, and anyway I don't see a way to make it work as it returns only `NameSet` and not actual values.
* `getMagmaLaws` which does it more properly (using `evalConstCheck`) crashes with `uncaught exception: unknown declaration`. This should never happen since the "check" part of the function name checks that the binding exists in the environment before evaluating. I tried several variations on this theme and couldn't find a way to get the values representing the laws.

In short, I'm out of my depth. Leaving this for somebody else to figure out.

amirlb:
disclaim

amirlb:
Looks like @0art0 fixed it a few hours ago.

I still have a branch with
* Output to stdout
* Adding a string form of the laws to the output `0 * 1 = 1 * 0` in addition to `{"lhs": {"left": {"leaf": 0}, "right": {"leaf": 1}}, "rhs": {"left": {"leaf": 1}, "right": {"leaf": 0}}}`
* Specifying a module to fetch the laws from
[...]

PR #530, opened 2024-10-12
use addAndCompile in equation command

digama0:
Fixes an issue [reported on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Using.20Environment.2EevalConst/near/476447645). Also adds support for namespaces to clean up a todo, although I don't think it's likely to matter.


vlad902:
Incidentally, is this a fix for #501 ?

amirlb:
I'm not 100% clear on the difference, but do we want to change the `addDecl` on line 130 too?

amirlb:
> Incidentally, is this a fix for #501 ?

I think it isn't, they types still won't match. But I have a ready-made solution for that after this PR is merged.

PR #532, opened 2024-10-12
graphiti: Add checkbox for sporadic equations

vlad902:
I noticed that sometimes graphiti can generate spurious vertices for sporadic equations. Add a checkbox to allow displaying them, otherwise explicitly filter them (in the correct way.)

Also, clean-up the generate graph script and trivially optimize the closure computation.


pitmonticone:
Thanks!

Issue #549, opened 2024-10-13
Zulip doesn't store conversations? Broken link in Tour of Selected equations


NooneAtAll3:
I've been browsing [Tour of Selected Equations](https://github.com/teorth/equational_theories/wiki/Tour-of-selected-equations), and the "Equations 65, 1491: "Asterix and Obelix"" paragraph in it has a following quote
> see [this ongoing discussion](https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/Equation.2065.20-.3E.20Equation.20359)"

link is dead because... Zulip doesn't seem to store old conversations?

(I'm not part of the project, but I've been participating in BBchallenge, so this kind of non-archiveability feels appalling)

But at the very least such dead links shouldn't stay public - so I suggest global search and prune of Zulip links in the repo (other than in Readme for the whole server, I suppose)

vlad902:
What I'm guessing happened here, is that topic was renamed and that the link therefore changed to [this](https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/Equation.2065.20-.3E.20left.20cancellability). I am not particularly familiar with Zulip, but googling around it looks like this is a known issue with Zulip being tracked [here](https://github.com/zulip/zulip/issues/21505) and that support for permalinks is coming shortly in the future, but this unfortunate behavior is the current state of things.

vlad902:
So reading that Zulip issue further, only links to topics are not currently permalinks, links to specific messages are. Links to topics end in /topic/NAME and links to messages end in /topic/NAME/near/NUMBER. I'll go ahead and open a PR shortly to update links to topics in the repo. I think only @teorth can update the wiki links.

teorth:
I've fixed that specific wiki link, thanks for pointing it out.

Chessing234:
PR #1444 adds the missing /near/ anchor to the Database of implications links in README and home_page/index.md — they were left as topic-only URLs after #554.

Chessing234:
Also opened #1445 for the remaining topic-only link in commentary/Equation879.md.

Chessing234:
Opened #1446 to add the missing permalink on the Equation677 finite-implication link.

Chessing234:
Opened #1447 for the Dupont/Dupond discussion link in Equation63 commentary.

Chessing234:
[...]

PR #554, opened 2024-10-13
Make zulip links permalinks (issue #549)

vlad902:
As brought up in issue #549, links to zulip topics are not permalinks [1] so a topic rename breaks all previous links to that topic. However, links to messages are permalinks, even after topic renames.

I've gone ahead and changed all the zulip links I found in the repo to permalinks to the first message in the linked threads to avoid future breakages.

[1] Yet. The zulip change to fix this is in development.


pitmonticone (approved):
Thanks!

PR #557, opened 2024-10-13
All unknowns 1701 -> ... resolved

zaklogician:
Resolves all 13 unknowns with antecedent 1701 negatively. Resolves 30 unknowns altogether.

Uses the modified translation-invariant magma construction, starting from pretty much the same base model as used for 1661 (see #511).

Refuting 1701 -> 4587 requires a fairly complicated countermodel of this sort. The other two countermodels are much simpler: 11 of the 13 implications did not even require patching beyond that required for truncation!

*Note: the dual proofs are not included, so the effective number settled by the construction is double that number. I'll create an issue for proving the duals once this is merged.*

### DIFF
--- a/equational_theories.lean
+++ b/equational_theories.lean
@@ -8,6 +8,7 @@ import equational_theories.Subgraph
 import equational_theories.Equations.All
 import equational_theories.InfModel
 import equational_theories.InfModel_1661
+import equational_theories.InfModel_1701
 import equational_theories.Generated
 import equational_theories.Counting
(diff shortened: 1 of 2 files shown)


teorth:
The transitive reduction algorithm now implements duality as part of the dashboard statistics, and we are nearly able to implement general metatheorems to convert one anti-implication to another (see https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/Generating.20non-implications.20for.20laws ), so having explicit dualizations of proofs may soon not be needed.

PR #560, opened 2024-10-13
Add button to hide fully-solved equations (#546)

### LINKED ISSUE #546: EQUATION_EXPLORER: Add option to hide fully solved equations in the main table
"Fully solved" means that there are 0 unknown equations that the equation either implies, or is implied by.

This will make it easier to search for interesting remaining equations, now that the number of outstanding implications has dropped to the mid hundreds.

vlad902:
And fix some minor bugs with the 'treat conjectures as unknown' buttons that I found along the way.

Closes #546


vlad902:
awaiting-review

PR #569, opened 2024-10-14
Formalises proof by Bernhard Rienke that 1648 implies 206 does not hold

Shreyas4991:
See the thread for the proof description and formalisation:
https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/1648.20!.3D.3E.20206/near/476775769

### DIFF
--- a/equational_theories/ManuallyProved/Equation1648.lean
+++ b/equational_theories/ManuallyProved/Equation1648.lean
@@ -11,15 +11,18 @@ theorem Equation1648_not_implies_Equation206 : ∃ (G: Type) (_: Magma G),
     op := fun x y => x - sign (y -x)
   }
   use ℤ, instMagmaInt
-  simp [Equation1648, Equation206]
+  simp only [Equation1648, Equation206, not_forall]
   constructor
   · intro x y
-    simp [instMagmaInt, sign]
+    simp only [sign, sub_neg, Int.reduceNeg, sub_sub_cancel_left, neg_eq_zero, ite_eq_left_iff,
+      sub_lt_self_iff, instMagmaInt]
     split <;> try simp_all
     case h.left.isFalse =>
-      split <;> simp_all <;> split <;> split <;> simp_all <;> linarith
+      split <;> split <;> split <;> simp_all <;> linarith
   · use 0,-1
-    simp [instMagmaInt, sign]
+    simp only [sign, sub_neg, Int.reduceNeg, sub_zero, neg_eq_zero, one_ne_zero, ↓reduceIte,
+      Left.neg_neg_iff, zero_lt_one, sub_neg_eq_add, zero_add, Int.reduceLT, zero_sub, sub_self,
+      zero_eq_neg, not_false_eq_true, instMagmaInt]
 
 theorem Equation1648_not_implies_Equation151 : ∃ (G: Type) (_: Magma G),
   Equation1648 G ∧ ¬ Equation151 G := by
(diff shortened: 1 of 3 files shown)


equational_theories/ManuallyProved/Equation1648.lean
```diff
@@ -19,3 +20,19 @@ theorem Equation1648_not_implies_Equation206 : ∃ (G: Type) (_: Magma G), Equat
...
+  Equation1648 G ∧ ¬ Equation151 G := by
+  let instMagmaInt : Magma ℤ := {
+    op := fun x y => if x > y then x + 1 else x - 1
+  }
+  use ℤ, instMagmaInt
+  simp [Equation1648, Equation151]
+
+  constructor
+  · intro x y
+    simp [instMagmaInt]
```
pitmonticone, line 34: Please use `simp only` here too.

equational_theories/ManuallyProved/Equation1648.lean
```diff
@@ -19,3 +20,19 @@ theorem Equation1648_not_implies_Equation206 : ∃ (G: Type) (_: Magma G), Equat
...
+    split <;> simp <;> split <;> simp_all <;> linarith
```
pitmonticone, line 35: Please use `simp only` here too.

equational_theories/ManuallyProved/Equation1648.lean
```diff
@@ -19,3 +20,19 @@ theorem Equation1648_not_implies_Equation206 : ∃ (G: Type) (_: Magma G), Equat
...
+  · simp [instMagmaInt]
```
pitmonticone, line 36: Please use `simp only` here too.

equational_theories/ManuallyProved/Equation1648.lean
```diff
@@ -19,3 +20,19 @@ theorem Equation1648_not_implies_Equation206 : ∃ (G: Type) (_: Magma G), Equat
...
   · use 0,-1
     simp [instMagmaInt, sign]
+
+theorem Equation1648_not_implies_Equation151 : ∃ (G: Type) (_: Magma G),
+  Equation1648 G ∧ ¬ Equation151 G := by
+  let instMagmaInt : Magma ℤ := {
+    op := fun x y => if x > y then x + 1 else x - 1
+  }
+  use ℤ, instMagmaInt
+  simp [Equation1648, Equation151]
```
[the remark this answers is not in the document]
Shreyas4991: These `simp` occurrences also apply other lemmas. It doesn't make sense to change it to `simp only`

equational_theories/ManuallyProved/Equation1648.lean
Shreyas4991: Same as above. `simp` is applying more useful simplifications here.

equational_theories/ManuallyProved/Equation1648.lean
Shreyas4991: Same as  above  : https://github.com/teorth/equational_theories/pull/569#discussion_r1799685118

equational_theories/ManuallyProved/Equation1648.lean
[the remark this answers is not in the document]
pitmonticone: I thought I should review as if we were in Mathlib. There are underlying technical reasons why `simp only` it's better. See [here](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/flexible.20vs.20rigid.20tactics) for further context.

Issue #506, opened 2024-10-11
COMBINATORIAL: Prove that 1076 does not imply 3


teorth:
See https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/1076.20!.3D.3E.203 for a proof.  The tasks are to

- Transfer Equation 1076 from `Eqns1000_1999.lean` to `Equations.lean` (leaving a commented out version in the former file);
- Add Equation 1076 to Chapter 2 of the blueprint;
- Add some version of the proof in the above discussion to the Asterix chapter of the blueprint (making a new section header if desired); 
- Formalize the statement and proof of the theorem in Lean (in some lean file TBD);
- Add the Lean file to the main `equational_theories.lean` file so that it is compiled by CI; and
- Add `\lean` and `\leanok` tags to the blueprint version of the proof.

madvorak:
claim

madvorak:
disclaim

madvorak:
I failed.

Here are some of my unsuccessful attempts:
```
import Mathlib


-- Daniel Weber thanks!
lemma FreeAbelianGroup.of_inj {α : Type} (a b : α) :
    FreeAbelianGroup.of a = FreeAbelianGroup.of b ↔ a = b :=
  ⟨(FreeAbelianGroup.of_injective ·), congr_arg _⟩


namespace PaceNielsen

abbrev A : Type := FreeAbelianGroup ℤ
abbrev p : A := FreeAbelianGroup.of 0
abbrev q : A := FreeAbelianGroup.of 1
abbrev r : A := FreeAbelianGroup.of 2
abbrev s : A := FreeAbelianGroup.of 3

notation "∅" => (none : WithZero A)
notation "O" => (some 0 : WithZero A)

noncomputable def E₀ : A →₀ WithZero A :=
  fun₀
  | 0 => p
  | p => q
  | -q => r
  | q + r => O
  | q + r - p => s
  | p - q - r + s => - q - r

def ℰ : Set (A →₀ WithZero A) := fun E =>
  ∀ a b c : A, (E a = b ∧ E b = c) →
    ∃ d : A, E (a - c) = d ∧ E (c + d - a) = -a

lemma E₀inℰ : E₀ ∈ ℰ := by
  have hpq : p ≠ q := by simp [FreeAbelianGroup.of_inj]
  have hpr : p ≠ r := by simp [FreeAbelianGroup.of_inj]
[...]

PR #577, opened 2024-10-14
[PARTIAL] Functional equational approach based on Pace Nielsen's proof

lyphyser:
I have formalized everything except the case 1 check and constructing the final function based on extensions, but don't have time to finish it, so posting this so someone can maybe finish it.

It's designed to constructively create the counterexample and represents the free abelian group using DFinsupp; it would be nice to cut down on that setup code, but I'm not sure if there is a shorter way of doing it especially constructively.

The partial sets are E are defined as partial functions, again using DFinsupp going to an Option type; the "fup" def lifts them to Option -> Option functions that are none on none.

The construction follows the paper and there are some additional notes on why Case 1 holds, which may help in designing its proof.


teorth:
I'm inclined to leave this PR unmerged for now until someone reclaims #506, and they can decide to what extent they will build off of this one.  (Though one argument in favor of merging now is that it will avoid merge conflicts and build issues later on.  So perhaps one can revisit this issue in a few days if nobody ends up reclaiming #506.)

Shreyas4991:
> I'm inclined to leave this PR unmerged for now until someone reclaims #506, and they can decide to what extent they will build off of this one. (Though one argument in favor of merging now is that it will avoid merge conflicts and build issues later on. So perhaps one can revisit this issue in a few days if nobody ends up reclaiming #506.)

In that case I will mark it as  a draft PR

teorth:
No longer required since 1076 has been formalized by other means.

PR #579, opened 2024-10-14
ManuallyProved: move InfModel_1XXX

zaklogician:
Moves the modified translation-invariant models refuting implications `1659 -> ...`, `1661 -> ...`, `1701 -> ...` to the `ManuallyProved` directory which will house these proofs in the future.

Some of these proves also get minor improvements (`Facts` instead of multiple theorems for counterexamples).

This is a refactor commit which changes equational_results. The output of `lake exe extract_implications outcomes --hist` was compared before/after the refactor to make sure they were identical and no proofs were lost.


Shreyas4991 (approved):
Looks good to me. I'll merge after CI passes

PR #580, opened 2024-10-14
Requirements.txt files and Python virtual environment setup scripts

### LINKED ISSUE #572: Add requirements.txt with package version numbers to assist in creation of Python environments
When working with Python scripts it is best practice to use a `requirements.txt` file so that all packages can be placed into one location and installed at once. As such a new user would be able to install all requirements through the use of one command
```
pip3 install -r requirements.txt
```
This also freezes packages into a consistent version that ensures that should the package be updated to a new, non-compatible version that our scripts will still work as intended. 

Documentation within the requirements file will also specify how to use it to install for users unaware with python package management practices

Joe-McCann:
Created requirements.txt files to keep all python dependencies in one place.

Normally all packages are thrown into a `requirements.txt` file and treated equally, regardless of whether they are project dependencies or sub-dependencies. The usage of `pip-config` to generate this requirements file allows us to see exactly where certain sub-dependencies are coming from. As such the logic is as follows

`requirements.in` lists off all python requirements that are imported within scripts explicitly. The command `pip-config` uses this file to generate the full `requirements.txt` file. 

_Note_: I have not included `pygraphviz` nor `leanblueprint` as of now in the `requirements.in` file as wasn't sure if that was necessary for local user setup. Could be discussed in comments 

_Second Note_: I have preemptively added `pandas` and `matplotlib` as project dependencies. As we move into stats and visualizations of data I feel these will be relevant and useful

---

Additionally setup scripts have been created for Unix and Windows environments to create virtual environments and activate them for new users, with some checks to make sure the user is utilizing them properly (ie correct directory). 

Closes #572


pitmonticone:
Thanks for your contribution @Joe-McCann. 

Regarding note 1: you just need to include `leanblueprint` which has `pygraphviz` as dependency.

Regarding note 2:  it's better not to add non-dependencies / unused libraries.

teorth:
A small request: can you update the "Scripts" section of CONTRIBUTING.md with a link to this new text file and some guidance for contributors?  Thanks!

Joe-McCann:
> Regarding note 1: you just need to include `leanblueprint` which has `pygraphviz` as dependency.

Yeah for `pygraphviz` I was concerned about this additional piece in specifying the build 
```
--global-option=build_ext --global-option="-L/usr/lib/graphviz/" --global-option="-R/usr/lib/graphviz/"
```
as I wasn't sure what the equivalent windows variation would be, and then it would just making dual compatibility a hassle. Making it work on the docker image + any user setup seemed like overkill so I assumed that if it wasn't needed for user devices to just leave it for the blueprint run. Can be changed if you think we should include

> Regarding note 2: it's better not to add non-dependencies / unused libraries.

Fair point! Will remove. 

> A small request: can you update the "Scripts" section of CONTRIBUTING.md with a link to this new text file and some guidance for contributors? Thanks!

Will do. Heading to teach soon so will do when I get back home tonight

Joe-McCann:
Updated PR addressing comments, let me know if there is anything else you'd like adjusted

pitmonticone:
LGTM. Thanks.

PR #585, opened 2024-10-15
Update blueprint action path filters

vlad902:
I noticed that after #582 landed the equation explorer dashboard didn't update. This should fix that and also add a missing filter for the graphiti data script.


pitmonticone:
We should probably remove the "home" path, right?

pitmonticone:
Thanks a lot! Merged.

PR #599, opened 2024-10-16
tree-based proof that 206 anti-implies 1648

### LINKED ISSUE #596: REGULAR_TREE: Formalize the tree-based proof that 206 anti-implies 1648
Details in https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/1648.20!.3D.3E.20206/near/476985846 .  This anti-implication will also be established by greedy algorithm methods, but the construction is visually appealing and may generalize to other equations.

The proof should go in the `ManuallyProved` section.

nomeata:
this fixes #596

Closes #596


nomeata:
I really shouldn't start working on fun little things at the evening. It’s 3am now…

I chose to made it self-contained, so there is 300 lines to develop the reduction function to construct the free product of three cyclic groups of order 2, and then 30 lines for the construction of the magma. Probably could have saved work by searching through mathlib… but where is the fun in that.

Will clean up tomorrow.

nomeata:
awaiting-review

nomeata:
I think it is, thanks for the ping.

PR #600, opened 2024-10-16
Rectifying requirements.txt

Joe-McCann:
`requirements.txt` was adjusted in a way that broke the creation of new virtual environments. 

In order to add new requirements, users should update the `requirements.in` file and then run the `uv` command at the top of that file. This will ensure that our `requirements.txt` file is as small as possible with no unnecessary dependencies


requirements.txt
Joe-McCann, line 1: Not sure why Github is displaying it like this, but you can see all top-level dependencies inside of the `.in` file

pitmonticone (approved):
Thanks

PR #614, opened 2024-10-16
chore: use UTF-8 instead of UTF-16LE for `requirements.txt`

euprunin:
```
% file requirements.txt
requirements.txt: Unicode text, UTF-16, little-endian text, with CRLF line terminators
% less requirements.txt
"requirements.txt" may be a binary file.  See it anyway?
<FF><FE>#^@ ^@T^@h^@i^@s^@ ^@f^@i^@l^@e^@ ^@w^@a^@s^@ ^@a
…
% dos2unix requirements.txt
dos2unix: converting UTF-16LE file requirements.txt to UTF-8 Unix format...
% less requirements.txt
# This file was autogenerated by uv via the following command:
…
```


pitmonticone (approved):
Thanks!

PR #618, opened 2024-10-17
a proof that our set of laws/equations is actually complete

nomeata:
https://leanprover.zulipchat.com/#narrow/channel/458659-Equational/topic/A.20final.20end-to-end.20theorem.20in.20Lean

### DIFF
--- a/equational_theories/Equations/LawsComplete.lean
+++ b/equational_theories/Equations/LawsComplete.lean
@@ -78,6 +78,36 @@ def Law.MagmaLaw.is_canonical (l : Law.MagmaLaw Nat) : Bool :=
   ((l.lhs.is_canonical 0).bind (fun n => l.rhs.is_canonical n)).isSome
 
 /--
+A decision procedure for checking a predicate for all canonical magma laws of a certain size.
+-/
+
+def testVars (n : Nat) (P : Nat → Nat → Bool) :=
+  (List.range n).all (fun i => P n i) && P (n+1) n
+
+def testAllSplits (s : Nat) (P : Nat → Nat → Bool) : Bool :=
+  (List.range (s+1)).all fun s' => P s' (s-s')
+
+partial def testFreeMagmas (s n : Nat) (P : Nat → FreeMagma Nat → Bool) :=
+  match s with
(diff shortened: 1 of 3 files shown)


amirlb:
From testing on my version of the code, looks like the comparison logic could be simpler: no need to compare expression size (can remove `.forks`), and no need for lots of conditions when comparing laws, this works for me:
```
def Law.MagmaLaw.comp (l1 l2 : Law.NatMagmaLaw) : Ordering :=
  (FreeMagma.comp l1.lhs l2.lhs).then (FreeMagma.comp l1.rhs l2.rhs)

def Law.MagmaLaw.is_canonical (law : Law.MagmaLaw Nat) : Bool :=
  law.symm.canonicalize.comp law ≠ .lt
```

amirlb:
Thinking about it some more, the fact the simple code works for comparison is probably because I generate only the canonical variable ordering in the first place. So fewer things need to be checked for filtering after that.

nomeata:
awaiting-review

pitmonticone:
Build failed with throwing the following error: 

```console
error: no such file or directory (error code: 2)
  file: ././././equational_theories/RArray.lean
Some required builds logged failures:
- Computing build jobs
error: build failed
```

nomeata:
classic, sorry

pitmonticone:
Thanks! I'll review in a few moments.

pitmonticone (approved):
Just quickly golfed a few bits. Thanks a lot! 

I'll merge after CIs

equational_theories/Equations/LawsComplete.lean
```diff
@@ -0,0 +1,334 @@
...
+        let mid := lb + w'
+        let l' := laws[mid]
+        if l.comp l' = .lt then
+          go lb w' fuel (by omega)
+        else
+          go mid (w-w') fuel (by omega)
+  termination_by structural fuel
+
+/-- The largest used variable. -/
+def FreeMagma.max : FreeMagma Nat → Nat
```
amirlb, line 86: I think all these '.max's aren't used

PR #636, opened 2024-10-18
represent verified implication graph in Lean

nomeata:
Work in progress. No closure taken yet.


nomeata:
The current design is infeasible as long as we have https://github.com/leanprover/lean4/issues/5771. But probably the graph is better represented as `Array Nat` or `RArray Nat`.

PR #638, opened 2024-10-18
clean-up: Eliminate try {}s

vlad902:
Marginally improves build times but more importantly makes the theorems far easier to understand.


pitmonticone:
Thanks @vlad902! Do we want to "touch" ATP generated proofs, @teorth? 

If so I can focus on golfing those in the next few days.

teorth:
We can, but then we may also want to update the code that generates those proofs as well, otherwise any edits will simply get wiped out if the code is run again.

pitmonticone:
Yes, exactly. 

@vlad902, can we guarantee that?

vlad902:
This is doable, I have a script that already reduces the number of `try {}`s in the workflow in the README. I will keep this in draft for now and get back to it in a few days.

pitmonticone:
Thank you very much @vlad902

PR #649, opened 2024-10-19
blueprint: minor corrections

vlad902:
In section 7.4.2 (Modified base models) I believe there is a minor error: the final corrected magma equation is a copy-and-paste of the previous equation that fails to satisfy Equation 1659. I fix it with the logic from the lean proof:

```math
x * y = \begin{cases} 0 & \text{if } x=0,\: x,y \text{ do not have the same parity} \\  x+1 & \text{if } x,y \text{ have the same parity} \\ x -1 & \text{otherwise} \end{cases}
```

I also fix one typo and a small clarification on the equation being referenced.


vlad902:
@zaklogician I think I got it right, but would you please verify the correction above?

pitmonticone (approved):

zaklogician:
BLUF: Your edit is correct. In fact, line `333` should get the same treatment, then this'll be ready to merge. Thanks a lot!

---

**I.** Let's think step-by-step. :)

I will use the verified Lean as reference.

```
  match x with
  | 0 =>
    if y % 2 = 0
    then 1 else 0
  | n + 1 =>
    if x % 2 = y % 2
    then n + 2 else n
```

This says that if $x=0$ and $y$ have the same parity, the result should be $1$. Otherwise, it should be $0$. So if 
$x$ and $y$ do not have the same parity, the result should be $0$ as your edit says, not the other way around as my original version says!

Yeah, your edited version is correct.

**II.** Line 333 and Line 366 should be the same. Line 333 is meant as signposting; basically I state that I will eventually modify the definition of Line 330 (which does satisfy both 1659 and 4315) to Line 333 (which still satisfies the former but not the latter). The rest is explanation about how one would come up with this particular modification.

vlad902:
> Line 333 and Line 366 should be the same. Line 333 is meant as signposting

Ah, I hadn't understood that while reading it. #699 is this last change.

PR #699, opened 2024-10-21
Add a blueprint correction from Zoltan

vlad902:
A correction from Zoltan as noted here https://github.com/teorth/equational_theories/pull/649#issuecomment-2423861379


pitmonticone (approved):
Thanks!

PR #702, opened 2024-10-21
Add support for explicitly Finite non-implications

vlad902:
As discussed in [1], I add support to the @[equational_result] attribute to parse non-implication theorems with Finite typeclasses. I then add support to `extract_implications` to specify `--finite-only` to limit results to finite non-implications.

In addition, I also add some additional checking to @[equational_result] as previously it would silently mis-parse theorems with the Finite typeclass.

This was tested by generating `extract_implications` closure/unknowns results before and after this change. When I included finite non-implications (not in this change), the results correctly showed them with `--finite-only`, while the results without `--finite-only` did not change and match the results prior to this change.

[1] https://leanprover.zulipchat.com/#narrow/channel/458659-Equational/topic/Proposal.3A.20Add.20Finite.20typeclasses.20to.20non-implications/near/477933508

### DIFF
--- a/equational_theories/Closure.lean
+++ b/equational_theories/Closure.lean
@@ -170,7 +170,7 @@ def equationSet (inp : Array EntryVariant) : Std.HashSet String := Id.run do
     | .implication ⟨lhs, rhs⟩ =>
       eqs := eqs.insert lhs
       eqs := eqs.insert rhs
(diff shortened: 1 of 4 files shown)


teorth:
Am merging this for now but one should do some further checks once more implications are converted to `Finite` to see if anything breaks.

PR #708, opened 2024-10-22
Codyroux/boolean ring

### LINKED ISSUE #354: BOOLEAN: Prove Theorem 5.8 (Sheffer stroke axiom)
The bare outline of a proof is sketched at https://teorth.github.io/equational_theories/blueprint/sect0004.html#sheffer , but one should refer to the paper https://www.cs.unm.edu/~mccune/papers/basax/v12.pdf for more details.  That proof is also computer assisted.

One should also locate the equation number for this equation (in principle this just requires running the `generate_eqs_list.py` script with `EQ_SIZE = 6`).

When done, mark the statement and proof of Theorem 5.8 with a `\leanok`, as well as a `\lean{}` tag.

codyroux:
Tie the knot from the excellent work from @jjtowery: prove that a certain messy equation is exactly what is needed for a Boolean Algebra. The proof is in 3 steps:

1. Prove that the complicated equation (equation 345169) implies the three original laws put forth by Sheffer in "A set of five independent postulates for Boolean algebras, with application to logical constants". Done by @jjtowery 
2. Prove that those 3 laws imply the structure of a "boolean ring", roughly following that paper. Done by myself.
3.  Prove that a "boolean ring" is a boolean algebra (in particular, define the order and prove that it is a nice lattice). Also done by myself, following the excellent wikipedia article: https://en.wikipedia.org/wiki/Boolean_algebra_(structure).

Closes #354


Shreyas4991:
Please use Mathlib's Boolean rings. I don't see why we need to re-invent that wheel

Shreyas4991:
awaiting-author

Shreyas4991:
> Ugh I could have sworn I checked. `BooleanRing` isn't referenced from `BooleanAlgebra` for some reason.
> 
> However the definition I wrote is "minimal" in the sense that it derives a number of things, including associativity, from a smaller set of equations. This is convenient when one is trying to prove that something is a BooleanRing, in particular a Magma satisfying Sheffer's axioms. Note also that the Mathlib one doesn't ask for commutativity (though it is easily derivable).
> 
> Worth noting that the join in my definition is just `. + .` rather than the (more standard) `x + y + x * y` from Mathlib. Both are used in different contexts, though the one in the Sheffer paper is the former.
> 
> However it's probably less work to show that the thing I defined is actually a Boolean ring proper than do the whole construction with `\leq`.
> 
> So what next? I could either:
> 
>     1. Just rename `BooleanRing` to something else, like `MinimalBooleanAlg` or something. I can do that today.
> 
>     2. Prove that this structure is actually a `BooleanRing`, (it is). I can't do this today.
> 
>     3. Eliminate the use of the intermediate structure, "inlining" the proof of associativity. This is a bit tedious and obscures the paper proof a bit IMHO, but whatever. Also can't do this today. Maybe Thursday?
> 
> 
> Thoughts?

I prefer option 3. I know it is tedious, but it minimises duplication. If you did keep the whole booleanring structure as is, you would still have a lot to explain regarding how this definition compares with Mathlib's and why we don't use it.

codyroux:
I take it "That's what Sheffer's paper and https://en.wikipedia.org/wiki/Boolean_algebra_(structure) do" is not a justification? I do agree that calling it a ring was a mistake, and probably the `+` and `*` symbols need to be renamed.

Shreyas4991:
At least in my opinion, it doesn't qualify. It would be an argument if  we were formalising Scheffer's paper. But I don't want to weigh in too much on this, since ultimately the goal here is to get a useful theory for some counterexamples.

For Mathlib, I am certain the maintainers will want a better reason. Bhavik already explained this on discord.

codyroux:
There's already 2 definitions though: a `BooleanAlgebra` and a `BooleanRing`! I could call this one a "Huntington algebra". Given my misfortunes though, I guess it would be wise to avoid a third structure.

I'll make the change when I find time.

Shreyas4991:
I discovered another reason to use Mathlib's definition. You get the `ring` tactic for free since BooleanRing extends Ring.

teorth:
Given that this theory is not essential for the completion of the primary goals of the project, I'm inclined to not rush the PR, but take the time to make it Mathlib-compatible, even if it means that the proofs then do not align perfectly with Sheffer's paper.

Presumably Mathlib won't care so much about the theory of equation 345169, but introducing a "Sheffer algebra" that is a magma obeying the three Sheffer axioms, and setting up a canonical equivalence between such algebras and `BooleanAlgebra` or `BooleanRing`, with various supporting API, sounds like something that Mathlib might like.  We can then have a separate set of results showing that 345169 is equivalent to the three Sheffer axioms (for non-empty magmas), which need not go into Mathlib.

codyroux:
All right! I have a rough idea of how to organize this, but it'll be a few days. If someone else wants to pick this up, ping me and I'll release it.

codyroux:
Here's where I am as of 10/27:

1. I refactored the code to have "Sheffer algebras" which encode the sheffer axioms and "Huntington algebras" which encode the Huntington axioms (which I previously and unfortunately called a Boolean ring). The join and meet operations are now appropriately named and noted.
3. I defined class constructors to move from Sheffer algebras to Boolean algebras via Huntington and Boolean algebras to sheffer algebras directly (it's a trivial proof).
4. Added an additional constructor for Boolean Algebras that takes in the axioms of a Huntington algebra.

Boolean rings did not enter the picture: the + operation defined in Mathlib is *not* the join of Huntington algebras, it is easier to just go directly to algebras.

What I did not do was eliminate the use of Huntington algebras: introducing variables instead of axioms does not seem to be an easy replacement in Lean, e.g. they can't be registered with the simplifier as far as I can tell and the variables need to be passed each time a lemma depending on them is called.

This may be a symptom of my inexperience with Lean.

At any rate I'm a bit exhausted and can't focus on this in the following weeks. I'm tempted to disclaim the issue and will come back to it if anyone shows interest.

jjtowery:
Hey @codyroux! So if I'm understanding the nature of the issue right now correctly, what we want for mathlib compatibility is to get rid of the Huntington algebra intermediary and instead go straight Sheffer algebra -> Boolean algebra, correct? I would definitely like to give a try at refactoring the proofs you've provided into that.
Something about this project has ended up leading to a lot of exhaustion, huh? I know the feel. :)

codyroux:
@jjtowery That is my understanding, yes though perhaps there may be other stylistic objections (e.g. I introduce a typeclass for the `x | y` notation, is that kosher?).

Please don't hesitate to make changes if you feel inspired!

jjtowery:
Hello @codyroux. Sorry for the delay but I finally got some time to work on it and I've managed to remove the Huntington algebra dependence by refactoring your proofs, and simplifying them here and there. It also made the `BooleanAlternate.lean` file unnecessary and just incorporated everything there into `ShefferAlgebra.lean`. Should I make a pull request on your branch to get it added to this commit?

codyroux:
Sure! Great job!

pitmonticone:
Thank you @codyroux and @jjtowery for your contributions! 

Could you please confirm that this PR is ready for review. At a quick glance seems good to me modulo some golfing of the proof code I can do later.

Shreyas4991:
Have we sorted out the issue of mathlib compatibility?

pitmonticone:
Thank you @codyroux and @jjtowery for your contributions! 

I have golfed and formatted `Sheffer.lean` and `ShefferAlgebra.lean`. 

I leave the rest of the review to @Shreyas4991.

codyroux:
Thanks a bunch @pitmonticone! Things are looking nice IMHO, but I'm not sure whether Mathlib has additional criteria I've overlooked.

jjtowery:
Speaking of mathlib compatibility, seems like mathlib decided to replace sup and inf with max and min, so I made a pull request to you @codyroux to fix that.
Other than this, the other 'not-mathlib-compatible' thing would be the theorem naming, but that's only important if we were actually trying to commit to mathlib. Otherwise, with @pitmonticone fixing the formattting, and that I fixed the issue of duplicating the BooleanRing definition, so there is no redundancy, this should be all good now.

codyroux:
Merged, let's see what the CI says.

Shreyas4991:
@codyroux : you were planning for some part of this to be PR'ed to mathlib right?

codyroux:
I certainly wasn't originally thinking this, but I guess it'd be nice. I'm looking at the [naming conventions](https://leanprover-community.github.io/contribute/naming.html) right now.

Shreyas4991:
@codyroux : I will wait for you to ping me here before reviewing/merging the LR

codyroux:
I've been on vacation, and sick, haven't forgotten about this but will need a bit more time.

codyroux:
@Shreyas4991 I'm ready!

codyroux:
@Shreyas4991 any updates?

Shreyas4991:
Hey sorry. I'm just returning from vacation. I'll look into it this week.

codyroux:
Oh, thanks a bunch!

equational_theories/Sheffer.lean
```diff
@@ -0,0 +1,162 @@
+import equational_theories.Generated.VampireProven.Sheffer
+import equational_theories.ShefferAlgebra
+
```
Shreyas4991, line 3: Do please write a module docstring for this file so that readers can get a quick overview of what this file proves

Shreyas4991 (approved):
Apart from the docstring I see no objections to merge this PR

equational_theories/Sheffer.lean
codyroux: Will do.

codyroux: Done!

codyroux:
😄

PR #723, opened 2024-10-23
Safe fixes generated by ruff

harahu:
I ran the [ruff python linter](https://docs.astral.sh/ruff/) on the repo. It generated the following (safe) automatic fixes.

One could add ruff to CI as well, but that would be more work.


pitmonticone (approved):
Thanks!

PR #732, opened 2024-10-25
Make "satisfies" in FME link to Graphiti

### LINKED ISSUE #728: FINITE_MAGMA_EXPLORER: Given a finite magma, return a graphiti depiction of all the laws it satisfies
I did this manually (with GPT help) in https://leanprover.zulipchat.com/#narrow/channel/458659-Equational/topic/854/near/478600628.2E01  but it might be useful to have this as a native feature.

Command-Master:
null

Closes #728

### DIFF
--- a/tools/fme/dist/elm.js
+++ b/tools/fme/dist/elm.js
@@ -5226,7 +5226,7 @@ var $author$project$Main$init = F3(
 		var maybeMagma = $author$project$Main$parseQuery(url);
 		var magma = A2($elm$core$Maybe$withDefault, '', maybeMagma);
 		return _Utils_Tuple2(
-			{K: url, L: key, M: '', B: $author$project$Main$ModelInfoNone, t: magma, Z: version},
+			{L: url, M: key, N: '', B: $author$project$Main$ModelInfoNone, t: magma, Z: version},
 			$elm$core$Platform$Cmd$none);
 	});
 var $elm$json$Json$Decode$string = _Json_decodeString;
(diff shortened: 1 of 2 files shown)


teorth:
Hmm, the feature hasn't shown up in the live FME yet as far as I can tell, but perhaps it requires some more time for the new build to be online.

Command-Master:
Perhaps I haven't built it correctly? @zaklogician , could you take a look?

Command-Master:
It looks like CI didn't run for some reason

zaklogician:
Looks like it has now landed in the live version, but it doesn't seem to work.

I'm still really busy atm, so can't assist, but if you can't resolve it, let me know or create an issue.
I should be able to take a look sometime next week.

zaklogician:
@Command-Master

I took a brief look just now and think I see two issues:

**1.** The url is `https://teorth.github.io/graphiti` and I think it should be `https://teorth.github.io/equational_theories/graphiti`.

**2.**

Instead of `a href`, this should go through an `onClick` event. Here's how `ExploreEquation` works when you click on an equation label:

```
    ExploreEquation num ->
      let
        base =
          "https://teorth.github.io/equational_theories/implications/?"
        url =
          base ++ String.fromInt num
      in
        (model, openInNewTab url)
```

You can add something like  `ExploreGraphiti (List (Int, String))` to `type Msg` and add a similar clause to the `update` function, which does what you currently do in `graphitiLink` to generate the `url`. Then replace your `a href` with `onClick (ExploreGraphiti (matching mi.satisfied))` and you'll be good to go.

You can also open index.html in `elm reactor` to test your changes locally.

Hope this helps.

PR #744, opened 2024-10-27
1485 !-> 2087,2124,3457,3511, full proof

digama0:
Finishes the greedy construction and proves the end-to-end theorems.


pitmonticone (approved):
Thanks a lot, Mario!

PR #617, opened 2024-10-17
BLUEPRINT: Unified chapter on infinite magma constructions (#591)

zaklogician:
**This PR:**

- Unifies the former Asterix Equation and Greedy Constructions chapters into a more comprehensive chapter titled "Infinite magma constructions".
- Adds a survey of examples, brief explanations of the following: a translation-invariant magma played straight, by Reinke to refute [1648->206](https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/1648.20!.3D.3E.20206/near/476775769); an example of my modified translation-invariant magmas, as written here: [1659->45](https://leanprover.zulipchat.com/#narrow/stream/458659-Equational/topic/Hard.20problems.20and.20negative.20results/near/476656060).
- Updates the selected equations in the blueprint and in the Lean code so that they remain in sync.

The surveyed examples are picked so that they are simple, but not trivial: the aim is not to showcase the most "powerful" applications of each technique, but to provide an accessible entrypoint for people who wish to study them.

**Review:**

This needs a brief review. Please read through 7.1 and 7.4 (the other two sections come from existing chapters), see that the web and LaTeX versions render OK, and that there are no math mistakes.

If that's done, it can and should be merged. Note that this progresses but does not by itself resolve #591 !

PR #758, opened 2024-10-30
add ref to asterix definition in section 7.2

dwrensha:
[section 7.2](https://teorth.github.io/equational_theories/blueprint/infinite-magma-constructions-chapter.html#asterix-section) mentions the "Asterix equation", but never recalls exactly what that equation is.

This PR adds a `\Cref` back to the [relevant definition](https://teorth.github.io/equational_theories/blueprint/subgraph-eq.html#eq65).

The style of the reference matches other in other chapters, e.g. https://github.com/teorth/equational_theories/blob/70e5481dbe5f931642966892c2cc7dda8cb1b02a/blueprint/src/chapter/constant.tex#L4

### DIFF
--- a/blueprint/src/chapter/infinite_magma_constructions.tex
+++ b/blueprint/src/chapter/infinite_magma_constructions.tex
@@ -38,7 +38,7 @@ For instance, writing $x = y+h$, we have
 $$ y \op x = x + f(h)$$
 $$ x \op (y \op x) = x + f(h) + f^2(h)$$
 $$ y \op (x \op (y \op x)) = x + f(h) + f^2(h) + f( h + f(h) + f^2(h) )$$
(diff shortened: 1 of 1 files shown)


dwrensha:
Digging through git history I made the following discovery:

The now-deleted file `asterix.tex` had the `\Cref`:

https://github.com/teorth/equational_theories/blob/88f806cecd84654bb4ced9a4426251c5fc65ebab/blueprint/src/chapter/asterix.tex#L11

but then the `\Cref` got removed when `asterix.tex` was folded into `infinite_magma_constructions.tex` by @zaklogician in #617.  I don't think there was a good reason for the removal, but maybe @zaklogician can correct me?

teorth:
Not sure why it was removed previously, but I see no harm in reinstating the \Cref.

PR #761, opened 2024-10-31
Relatively definable equations

Timeroot:
Along the lines of what's described [here](https://leanprover.zulipchat.com/#narrow/channel/458659-Equational/topic/Future.20directions/near/479506945). I'm defining four reasonable (imho) ways to talk about one equation implying the structure of another equation on that same magma.
 * "Definable" means that, if I have an operation * obeying law L, then there's a FO formula that defines the graph of an operation + that obeys the other law.
 * It's "Term-Definable" if this can be done with just a term, and expression, like `x+y = x*((x*y)*(y*x))`. I think this name exists in the literature, but it seems rare, and I am not a model theorist.
 * As Will Sawin pointed out, there's a stronger property which is that you can then recover the original operation from the new one. This is like saying each is definable from the other, but is more than that, because they have to actually be the same operations (just being definable from each other, they could each be used to construct a 'boring' operation.) I'm calling this "Structural".
 * And, "Term-Structural" as the analogous.

There are implications (which I've proved in the Lean) that the equational_theories "implies" leads to term-structural; and then term-structural leads to structural leads to definable, and term-structural leads to term-definable leads to definable.

Tarski's axiom is the strongest relationship that isn't an actual implication, it's term-structural.

I've got the fact that term-* relations are transitive, need to prove transitivity for the other two. Then I'll prove reflexivity, and that's enough to show we've got 5 different Hasse diagrams, adding more collapses. And then I'll actually prove a few examples and non-examples.

Some of this definitely should be upstreamed into Mathlib later btw, at least `LEquiv.onTerm` and `Functions.term`. When I prove the transitivity of definability, that might be something too.


Timeroot:
Got an example of the "main" result in, that Law543 implies a commutative group structure, so in particular we get a term-definable associative law, and a commutative law.

...The associative law is actually trivial, since `x ◇ y := x` means that there is always a term-definable associative law, as Will Sawin pointed out. But the commutative result is nontrivial, and I've got most of the boilerplate I need to start tackling the "structural" nature, that there is an associative law that _lets you recover the original operation_.

PR #766, opened 2024-10-31
Add a "Show Proof" button to the equation explorer.

### LINKED ISSUE #756: EQUATION EXPLORER: Integrate the #741 script
We have a script at #741 to display the details of any given implication or anti-implication, assuming that a suitable JSON file has been built.  One could integrate this functionality into EE, for instance by adding a "Proof" link to any implication, anti-implication, or equivalence.

amirlb:
There is a back-end change to the @equational_result tag, to save the line number as well as the file name.

This allows the front-end (in a new show_proofs.html file) to link directly to the location of the proof. It would probably be more elegant to show this as a pop-up, but I'm not familiar enough with web development to do this easily.

Closes #756


amirlb:
Tagging @carlini

amirlb:
I mixed up the order of non-implications. Implications show in the correct order, but A ⇏ B is displayed as
`B ⇒ ... ⇒ C ⇏ D ⇒ ... ⇒ A`, while the correct non-implication is `D ⇏ C`.

I'm still debating what to show for non-implications, currently my idea is to show `A ⇐ ... ⇐ D ⇏ C ⇐ ... ⇐ B`.

The page lists everything vertically, so it will mean up arrows for implications and down-arrows for non-implications. An example of the bug:
image

amirlb:
I think a more elegant approach would be to tweak horizontal spacing to get
```
        B
        ⇓
        ...
        ⇓
D   ⇏   C
⇓
...
⇓
A
```
but it's a bit hard.

amirlb:
awaiting-review

amirlb:
Another question is how the added data impacts load time. I removed some whitespace from the data file, which makes it smaller even though all the source locations were added. But the compressed file size is ~50% larger.

I don't know how to measure this before deploying, since it depends on the server too, and I don't know how to reproduce Github Pages' choices locally.

amirlb:
I also have a small code change to auto-focus the text field and to have Enter jump to the equation page. Is it considered good form to add this to this branch?

teorth:
Sure, may as well bundle these small improvements together.

teorth:
I'm going to experimentally merge this to see how it actually works "live".

PR #769, opened 2024-11-01
prove duality theorems

digama0:
The command `duals 1485 ↔ 2162` proves a theorem `duals_1485 : IsDual Law1485 Law2162`. Statements like this are generated for every dual pair of equations (including self-duals).


digama0:
I think you might have merged too quickly, the build is still running and it looks like it won't end well. I'll push another PR with the fixes

PR #785, opened 2024-11-03
Bol/Moufang laws

Timeroot:
See https://leanprover.zulipchat.com/#narrow/channel/458659-Equational/topic/Bol.2FMoufang.20loops/near/480277948


Timeroot:
I added the facts that these are all implied by 4512 (associativity) and are not implied by equation 3 (x = x ◇ x) or 43 (commutativity). But I'm hoping that some files in `Generated` can include these in the future? :)

PR #789, opened 2024-11-03
prove LawsComplete optimization

digama0:
I managed to prove the completeness of the canonicalization mechanism with respect to the original, optimized canonicity property, so we can restore the optimization removed in #788 .

### DIFF
--- a/equational_theories/Equations/LawsComplete.lean
+++ b/equational_theories/Equations/LawsComplete.lean
@@ -211,7 +211,7 @@ where
     | some i => let (l', xs) := go l xs; (i::l', xs)
     | none => let (l', xs') := go l (xs ++ [v]); (xs.length :: l', xs')
 
-def List.canonicalize.go_append {α} [DecidableEq α]
+theorem List.canonicalize.go_append {α} [DecidableEq α]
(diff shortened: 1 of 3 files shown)


equational_theories/Equations/LawsComplete.lean
```diff
@@ -104,27 +190,43 @@ def Law.MagmaLaw.max (l : Law.MagmaLaw Nat) : Nat := Nat.max l.lhs.max l.rhs.max
...
+  (go l []).1
+where
+  go : List α → List α → List Nat × List α
+  | [], xs => ([], xs)
+  | v :: l, xs =>
+    match xs.indexOf? v with
+    | some i => let (l', xs) := go l xs; (i::l', xs)
+    | none => let (l', xs') := go l (xs ++ [v]); (xs.length :: l', xs')
+
+def List.canonicalize.go_append {α} [DecidableEq α]
```
amirlb, line 214: should this `def` be a `theorem`?

digama0:
```suggestion
theorem List.canonicalize.go_append {α} [DecidableEq α]
```

pitmonticone (approved):
Thanks!

PR #830, opened 2024-11-13
Translation-invariant greedy construction for 879 =/=> 4065

### LINKED ISSUE #633: MODIFIED_GREEDY: Refute 879 -> 4065
See [this discussion](https://leanprover.zulipchat.com/#narrow/channel/458659-Equational/topic/Equation.20879.20!.3D.3E.204065) for the proof.  

The formal proof can go in the `ManuallyProved` folder. `ManuallyProved.lean` should be updated afterwards, and the appropriate conjectures from `Conjectures.lean` removed.

amirlb:
null

Closes #633


teorth:
I take it from the sorries that this is still a work in progress?  If so, I can mark it as a draft.

amirlb:
Yes, it's still I'm progress. Changed to draft.

amirlb:
Hi, updating that I've been busy for a week and a bit, and am planning to go back to this and wrap it up in the next few days. Probably using @b-reinke 's constructions, there are lots of examples to learn from, and the proofs turn out so short.

teorth:
@b-reinke's arguments use a non-translation-invariant construction rather than a translation-invariant one.  If you want to switch to the former, we can try looking for a non-translation-invariant construction for 879 along the lines of the ones in https://teorth.github.io/equational_theories/blueprint/infinite-magma-constructions-chapter.html#greedy-section , but they might not be compatible with the code you have so far, so we didn't try to make such a construction yet. Anyway, let me know if you'd like such a construction.

amirlb:
I mean adapting his formalization of existing translation-invariant proofs (eqs 73, 719, 1518, 1526, 1722), not finding a different proof

amirlb:
Reading back on the issues here and on the post, it seems I don't have the right terminology. For instance, equation 73 uses the `GreedyMagma` whose operation is
```
noncomputable instance instMagma : Magma (GreedyMagma e₀) where
  op x y:= op e₀ x y
```
with
```
def op (e₀ : Extension) (x y : A) : A := f e₀ (y*x⁻¹) * x
```
I don't see why it isn't translation-invariant, since `op x (a * x) = f a * x`.

In any case, this is the style I intend to copy.

teorth:
Ah, I didn't realize that many of Bernhard's proofs were still based on the translation-invariant models.  You can ignore my previous comment then!

Issue #809, opened 2024-11-09
FINITE_IMPLICATIONS: Prove Lemmas 3.7 and 3.8


teorth:
[Lemma 3.7](https://teorth.github.io/equational_theories/blueprint/infinite-model-chapter.html#ffg) and [Lemma 3.8](https://teorth.github.io/equational_theories/blueprint/infinite-model-chapter.html#gff) are lemmas about functions on finite sets that can prove implications in finite magmas, see https://leanprover.zulipchat.com/#narrow/channel/458659-Equational/topic/Austin.20pairs/near/480601897

When done, place \lean and \leanok tags in the blueprint accordingly.

vlad902:
claim

vlad902:
propose PR #931

PR #846, opened 2024-11-18
Formalize finite implications from 3342

vlad902:
Two small notes to the reviewers:
- This is my first time using `\lean` and `\leanok` in the blueprint, it seems straightforward but please check that I'm using it correctly.
- `Finite.fn_eventually_periodic` (and maybe `fn_periodic`) will probably be moved to somewhere common for use in #809, I included it with the proofs for 3342 for now since I'm not done with that proof yet and not sure what it will look like in the end.

### DIFF
--- a/blueprint/src/chapter/infinite_models.tex
+++ b/blueprint/src/chapter/infinite_models.tex
@@ -128,18 +128,20 @@ Many implications for finite magmas rely on the fact that if $f, g:X \to X$ are 
 This can be used to obtain a few positive finite magma implications, for instance by setting $f,g$ to be left and right multiplication operators.  Another useful lemma is
 
 \begin{lemma}[Eventual period]\label{period}  Let $X$ be finite and $f: X \to X$.  Then there exists $n \geq 1$ such that $f^{2n} = f^n$.
+  \lean{Eq3342.Finite.fn_periodic}\leanok
 \end{lemma}
 
-\begin{proof}  By the pigeonhole principle, there exists $n \geq 1$, $m \geq 0$ such that $f^{m+n} = f^m$, which implies on iteration that $f^{m+mn} = f^m$ and hence $f^{2mn} = f^{mn}$, giving the claim.
+\begin{proof} \leanok By the pigeonhole principle, there exists $n \geq 1$, $m \geq 0$ such that $f^{m+n} = f^m$, which implies on iteration that $f^{m+mn} = f^m$ and hence $f^{2mn} = f^{mn}$, giving the claim.
 \end{proof}
 
 As a sample application, we have
 
 \begin{corollary}[3342]\label{3342}  On a finite magma $M$, equation 3342, $x \op y= y \op (x \op (x \op x))$, implies
+  \lean{Eq3342.Finite.Equation3342_implies_Equation3522,Eq3342.Finite.Equation3342_implies_Equation4118}\leanok
 equation 3522, $x \op y = x \op ((y \op y) \op y)$, as well as equation 4118, $x \op y = ((x \op x) \op x) \op y$.
 \end{corollary}
 
-\begin{proof}\uses{period}  Write $Sx := x \op$, $fx := x \op Sx$ and $Cx = Sx \op x$, then we have $x \op y= y \op fx$, and our task is to show $x \op y = Cx \op y$ and $x \op y = x \op Cy$ for finite magmas, giving the four open implications.
+\begin{proof}\uses{period} \leanok Write $Sx := x \op$, $fx := x \op Sx$ and $Cx = Sx \op x$, then we have $x \op y= y \op fx$, and our task is to show $x \op y = Cx \op y$ and $x \op y = x \op Cy$ for finite magmas, giving the four open implications.
 
 Note that $x \op y= y \op fx = fx \op fy$, hence $f$ is a homomorphism.  By \Cref{period}, there is $n \geq 1$ with $f^n = f^{2n}$.  From 3342 we have $Sx = S fx$ and $Cx = Sx \op x = f^n Sx \op f^n x = f^n Sx \op f^{2n} x = f^{2n-1} x \op f^n Sx = f^{n-1} x \op S x = f(f^{n-1} x) = f^n x$.  Then
 $$ Cx \op y = f^n x \op y = f^{2n} x \op f^n y = f^n x \op f^n y = x \op y$$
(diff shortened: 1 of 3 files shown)


equational_theories/ManuallyProved/Equation3342.lean
```diff
@@ -58,4 +63,106 @@ theorem Equation3342_facts : ∃ (G : Type) (_ : Magma G), Facts G [3342] [3456,
...
+    refine ⟨p, hpgt, this⟩
+  . let n := s * p
+    have : f^[n] = f^[2*n] := by
+      unfold n
+      obtain ⟨ppred, hppred⟩ := Nat.exists_eq_succ_of_ne_zero (by linarith)
+      rw [hppred, Nat.succ_eq_add_one, Nat.mul_succ, Nat.add_comm]
+      have : 2 * (s + s * ppred) = s + s * ppred + s * p := by simp_arith only [hppred,  Nat.mul_succ]
+      rw [this, ← hmod]
+    have ngt : n > 0 := by apply Nat.mul_pos h hpgt
+    refine ⟨n, ngt, this⟩
```
madvorak, line 100:
```suggestion
    exact ⟨n, ngt, this⟩
```

equational_theories/ManuallyProved/Equation3342.lean
```diff
@@ -58,4 +63,106 @@ theorem Equation3342_facts : ∃ (G : Type) (_ : Magma G), Facts G [3342] [3456,
...
+
+lemma main_result (G : Type*) [Magma G] [Finite G] (h : Equation3342 G) : Equation3522 G ∧ Equation4118 G := by
```
madvorak, line 102: This is a weird decomposition into lemmas, but I can live with it.

equational_theories/ManuallyProved/Equation3342.lean
```diff
@@ -58,4 +63,106 @@ theorem Equation3342_facts : ∃ (G : Type) (_ : Magma G), Facts G [3342] [3456,
     use (0, 0), (0, 0)
     decide
 
+-- https://leanprover.zulipchat.com/#narrow/channel/458659-Equational/topic/Austin.20pairs/near/482525422
+
+lemma Finite.fn_periodic (G : Type*) [Finite G] (f : G -> G) : ∃ p : ℕ, p > 0 ∧ f^[p] = f^[2*p] := by
+  have Finite.fn_eventually_periodic : ∃ s p : ℕ, p > 0 ∧ f^[s] = f^[s+p] := by
+    obtain ⟨p1, p2, lt, heq⟩ : ∃ p1 p2 : ℕ, p1 < p2 ∧ f^[p1] = f^[p2] := by
```
madvorak, line 70: Write lower indices please.

\1
\2

madvorak:
Congratulations!

equational_theories/ManuallyProved/Equation3342.lean
vlad902: If you mean that there are two lemmas (fn_periodic and main_result), the first lemma will probably be reused in a future proof so it was easier to keep it separate. If you mean the fact that this lemma proves both equations, I agree that it's weird, but trying to construct it so the equation-specific proofs were in the separate implications led to requiring some of the internal hypotheses to be split out and that led to thers needing to be separate and it seemed worse than this approach in the end, but I'm a Lean-beginner and very open to feedback on how to structure it better!

vlad902:
Thank you for the feedback!

equational_theories/ManuallyProved/Equation3342.lean
madvorak: You know better has to structure your proof than I do.

I just know the common pattern — when two proofs share a lot of code, it is better to extract the shared part into a lemma than to merge the two theorems in order to have the shared parts only once. You can also use ˙private lemma˙ if you want.

equational_theories/ManuallyProved/Equation3342.lean
```diff
@@ -58,4 +63,107 @@ theorem Equation3342_facts : ∃ (G : Type) (_ : Magma G), Facts G [3342] [3456,
...
+    have : f^[n] = f^[2*n] := by
+      unfold n
+      obtain ⟨ppred, hppred⟩ := Nat.exists_eq_succ_of_ne_zero (by linarith)
+      rw [hppred, Nat.succ_eq_add_one, Nat.mul_succ, Nat.add_comm]
+      have : 2 * (s + s * ppred) = s + s * ppred + s * p := by simp_arith only [hppred,  Nat.mul_succ]
+      rw [this, ← hmod]
+    have ngt : n > 0 := by apply Nat.mul_pos h hpgt
+    exact ⟨n, ngt, this⟩
+
+private lemma main_result (G : Type*) [Magma G] [Finite G] (h : Equation3342 G) :
```
madvorak, line 102: What I meant is that, if you wanted, you could make `private lemma` out of those `have` inside this proof and then refer to them in both theorems.

I understand if you don't feel like refactoring it. In this case, I'd just change `private lemma main_result` to `private theorem main_result` and merge it soon.

vlad902: Yeah, I just gave it a go and gave up again. You need to pass around the period and related hypotheses and it made it less clear in the end, even if it was more modular. I've just made the lemma->theorem change you recommend.

madvorak:
LGTM, somebody just check the `leanok` which I don't know how to use either.

blueprint/src/chapter/infinite_models.tex
```diff
@@ -128,18 +128,20 @@ \chapter{Infinite models}\label{infinite-model-chapter}
 This can be used to obtain a few positive finite magma implications, for instance by setting $f,g$ to be left and right multiplication operators.  Another useful lemma is
 
 \begin{lemma}[Eventual period]\label{period}  Let $X$ be finite and $f: X \to X$.  Then there exists $n \geq 1$ such that $f^{2n} = f^n$.
+  \lean{Eq3342.Finite.fn_periodic}\leanok
```
pitmonticone, line 131:
```suggestion
\begin{lemma}[Eventual period]\label{period} \lean{Eq3342.Finite.fn_periodic} \leanok
    Let $X$ be finite and $f: X \to X$.  Then there exists $n \geq 1$ such that $f^{2n} = f^n$.
```

vlad902:
I've made the mentioned updates but I've been stuck making the build work. My limited understanding is that the recent mathlib/lean bump is responsible for it, but I don't fully understand it. Previously, e.g. in the following example:
```
example {G : Type*} [Finite G] (f : G → G) : False := by
  obtain ⟨p₁, p₂, ne, heq⟩ := Finite.exists_ne_map_eq_of_infinite (Nat.iterate f ·)
```
[Finite G -> G] was automatically synthesized, and now it's not. Perhaps this is due to lean4#6024, I'm trying to debug.

vlad902:
The cause was much more prosaic: I believe mathlib 9a9cb1894400637ebac1c1c8ac2702c23412ee5e simply split some imports I was previously transitively depending on, and I no longer imported the instance I needed.

PR #933, opened 2024-11-27
Change the freshGenerator functions to allow free groups with arbitrary countable set of generators

amirlb:
See discussion around https://leanprover.zulipchat.com/#narrow/channel/458659-Equational/topic/Formalising.20Counterexamples.20from.20the.20Greedy.20Method/near/484653113


amirlb:
Sure, it looks like an easy fix - it's in code that was duplicate of FreshGenerator.lean which I removed. I'll resolve the conflict, but I prefer to do it at my computer to make sure the code doesn't need to be updated after the version bump. So in a few hours :)

pitmonticone:
Thank you!

PR #963, opened 2024-11-30
Add the original Austin law

vlad902:
Just for fun, I added Austin's original law from [1], and also included the proof that it has no finite models. The proof that it has satisfying infinite models is not done.

[1] A. K. Austin. A note on models of identities. Proc. Amer. Math. Soc., 16:522–523, 1965.


teorth:
Are you planning to add the infinite model case to this PR in due course, or would you prefer that this PR be merged as is?

vlad902:
I think it should be merged now. I'm not sure if I'll get to the infinite case, and I think it's a nice historical identity to have regardless of that proof.

PR #891, opened 2024-11-23
[CONTRIBUTIONS]: Pietro Monticone 

pitmonticone:
- [ ] **Conceptualization**
- [x] **Data Curation**: 
  - I wrote, reviewed, edited the blueprint (see [here](https://github.com/teorth/equational_theories/commits/main/blueprint?author=pitmonticone)).
  - I added the edge list of the implication graph and the script to generate it (see [here](https://github.com/teorth/equational_theories/commits/main/data?author=pitmonticone)).
- [x] **Formal Analysis**: 
  - I explored Directed Link Prediction (DLP) on the implication graph using Graph Neural Network (GNN) autoencoders to predict missing edges and estimate the truth value of unproven implications. A detailed report, including insights from test sets on edge existence, directionality, and bidirectionality, is available [here](https://leanprover.zulipchat.com/#narrow/channel/458659-Equational/topic/Graph.20ML.3A.20Directed.20link.20prediction.20on.20the.20implication.20graph).
- [ ] **Funding Acquisition** 
- [ ] **Investigation**
- [ ] **Methodology** 
- [x] **Project Administration**: 
  - I set up, tested, maintained and optimised the project environment and workflows (see [here](https://github.com/teorth/equational_theories/commits/main/.vscode), [here](https://github.com/teorth/equational_theories/projects) and [here](https://github.com/teorth/equational_theories/commits/main/.github?author=pitmonticone)).
  - I coordinated with other maintainers and contributors (see [here](https://leanprover.zulipchat.com/#narrow/channel/458659-Equational))
  - I have managed and continue to manage the project's dependency updates (see [here](https://github.com/teorth/equational_theories/pulls?q=author%3Apitmonticone+bump)). 
- [x] **Resources**:
  - I contributed to the development of the [LeanBlueprint](https://github.com/PatrickMassot/leanblueprint) library used to generate the project blueprint (see [here](https://github.com/PatrickMassot/leanblueprint/commits/master/?author=pitmonticone)). 
  - I created and currently maintain the [LeanProject](https://github.com/pitmonticone/LeanProject) template used to generate this project (see [here](https://github.com/pitmonticone/LeanProject/commits/main/?author=pitmonticone)). 
- [x] **Software**: 
[...]

PR #1003, opened 2024-12-09
Add numbers to paper text for the 4x4 magma search

carlini:
Adds the numbers to the paper text for the 3x3 and 4x4 magma brute force run. It resolves 13,632,566 implications with 524 distinct magmas.


carlini:
I basically disappeared for the last two months and it looks like you all are trying to make this into a paper, how exciting!

I don't know how people write math papers and so don't know how much detail to put here about the implementation so I've just left it all out. If it would be worth including some explanation of what's going on for this search I can write that, but it's not super exciting. 

(It does take a little bit of work because while it's "just" 10^9 equations tables, you have to run each of these over 4000 equations and checking each equation requires checking every assignment of variables. Doing this brute force would mean ~7*10^16 work.)

I also don't know how you all are handling authorship. I didn't do any of the interesting math so I've left my name off for now.

teorth:
Thanks for the contribution!

Regarding authorship, following the discussion at https://leanprover.zulipchat.com/#narrow/channel/458659-Equational/topic/Drafting.20the.20paper ,what we are doing is asking everyone who has made a non-trivial contribution to the project to submit a PR adding an entry for themselves in https://github.com/teorth/equational_theories/blob/main/paper/contributions.md with some brief commentary in the PR to explain the contributions.  We plan to have quite a broad set of authors in the end.  You can see some examples of previous PRs at #974 #913 #896 #891 #884.  Please feel free to submit such a PR, your contributions were certainly valuable!

PR #1011, opened 2024-12-10
Prove-1133-implications

### LINKED ISSUE #985: FINITE: Establish 1113->1167->1096
Convert the conjectures at https://github.com/teorth/equational_theories/blob/main/equational_theories/ManuallyProved/Equation1113.lean to theorems, following the blueprint proofs.

When done, add `\lean` and `\leanok` tags to the blueprint as appropriate.

teorth:
null

Closes #985


vlad902:
Could we rename this file and it's namespace to be 1133 instead of 1113

PR #1032, opened 2024-12-29
Proofs of equations that are/are not full spectrum

vlad902:
As noted in [1], these few proofs should by transitivity/duality prove the full-spectrum/not-full-spectrum property for all equations of order 4.

[1] https://leanprover.zulipchat.com/#narrow/channel/458659-Equational/topic/Equations.20with.20full.20spectrum


dsm054:
@vlad902: maybe the docstring could be tweaked regarding the non-full implication?

vlad902:
Can you clarify what change you mean?

dsm054:
Ha, my mistake: on first read I didn't see you'd flipped the direction on "All equations that are implied by a full spectrum must be full spectrum and all equations that imply a not-full-spectrum equation, must not be full spectrum" (cf. comment on Zulip).

PR #1080, opened 2025-02-15
Change scripts to use JSON output from export_equations

guangyusong:
Updates the extraction process to read from full_entries.json (produced by export_equations)

Removed the Batteries imports since Lean’s core now covers the required functionality.


vlad902:
I don't think we want this change: it does not update in the future once we add new implications so extract_implications now no longer gets new updates. In fact `full_entries.json` is outdated, so the dashboard is now broken.

teorth:
Ah, I did not appreciate that the dashboard did not run through the JSON output.  (If the project still had some ways go to, I would recommend documenting these sorts of back-end architectures, but I guess we soon won't need to update the dashboard much further, so the issue is largely moot.)

PR #1152, opened 2025-04-03
reduce injectivity of L_0 embedding to simpler claims

teorth:


teorth:
Merging for now to lock in a significant number of refactorings to head off future merge conflicts.

PR #1160, opened 2025-04-11
update quasi-primal to idemprimal

teorth:


kotenok2000:
Line  "It was also shown in \cite{murskii-2} that ``almost all'' magmas $M$ (in a certain precise sense) are" is duplicated in that commit.

teorth:
Thanks, I'll fix it now

PR #1188, opened 2025-04-28
Update intro.tex

snburris:
This corrects the coefficient of the formula for E(a,b) when a is not equal to b Also it replaces "idempotent bijection" with the correct word "involution", and replaces the function name "Idem" with "Invol"


snburris:
An updated LaTeX file of the counting formula will soon be submitted for inclusion in the Scripts folder on counting magma equations.

PR #1195, opened 2025-05-05
Fixing Alternate counting formula

snburris:
This fixes the following:
(1) change idempotent to involution
(2) change multline* to align*
(3) change lead coeff of formula for E(a,b) when a \neq b


teorth:
Unfortunately I wasn't able to fix this one issue in the web version of the blueprint, although the print version at https://teorth.github.io/equational_theories/blueprint.pdf works just fine, and I otherwise see no compilation errors.   Possibly this issue will magically fix itself with subsequent updates of the blueprint (these bugs can be weird that way).

snburris:
Thanks for trying to sort this out and making the .pdf version of the blueprint where indeed everything looks perfect! Since the expression for E(a,b) when a \neq b is fine in the blueprint, but there are problems with the case that a = b > 0, I have to wonder if the web version has trouble with arrays. The parts of LaTeX that appear in the latter E(a,b) formula are essentially the same as in the former except that the latter has the formula spread over 3 lines, whereas the former has a single line using $$...$$. 
I right-clicked on the error message in the blueprint and then on the "show Tex commands" and could see the insertion of the text "bgroup" in the LaTeX code---no idea where that came from. I am not tech-savy enough to chase this down, but it is likely that others will run into this problem and it will be fixed. I don't know how you find the time to look at these small issues when there is the big picture of getting the project wrapped up so a final paper is available. But I do appreciate your attention to such details!
Best,Stan

    On Tuesday, May 6, 2025 at 04:54:22 p.m. EDT, teorth ***@***.***> wrote:  
 
 teorth left a comment (teorth/equational_theories#1195)
Unfortunately I wasn't able to fix this one issue in the web version of the blueprint, although the print version at https://teorth.github.io/equational_theories/blueprint.pdf works just fine, and I otherwise see no compilation errors. Possibly this issue will magically fix itself with subsequent updates of the blueprint (these bugs can be weird that way).

pitmonticone:
Fixed #1198

PR #1198, opened 2025-05-06
[BLUEPRINT]: Fix "Unknown environment `bgroup`" issue

pitmonticone:


snburris:
Fantastic! I won't ask what you did because I probably wouldn't understand it. But I am impressed!

PR #1208, opened 2025-05-12
BB5 paper citation

tcosmo:
The BB5 paper has reached its pre-print stage (see https://github.com/bbchallenge/bbchallenge-paper/issues/9), which should be on arxiv by the end of the summer. 

The title and author list are stable, hence I added a bibtex entry and citation in the intro of the ETP paper where Busy Beaver is mentioned. I also added a bibtex entry and citation for Coq 8.20.0 which is the Coq version that we used (just before it became Rocq with v9).

Thank you very much,


teorth:
Thanks!

PR #1219, opened 2025-05-16
add commentary

teorth:


blefloch:
I thought there was a bug in the equation explorer, but actually these are in the end just forms of Equation 4.  The commentary should mention that, please.

teorth:
I've updated the commentary file accordingly.

Issue #1218, opened 2025-05-14
FRONTEND: Get equation explorer to mention canonically equivalent form


teorth:
Proposed in https://leanprover.zulipchat.com/#narrow/channel/458659-Equational/topic/Questions.20for.20equation.20commentaries/near/518122864 : if viewing an equation in Equation Explorer that is equivalent to a lower numbered equation, mention the lowest numbered equivalent equation near the top of the page (in addition to where it is already listed at the end of the page).  Furthermore, if this "canonically equivalent form" already has a commentary file, display that file (in addition to any file present for the non-canonical form).

blefloch:
Ideally it would be best to also account for the dual equation and the corresponding commentary.  The logic could be:
1. Display equation number, and lower-numbered equivalent equation if applicable (currently not done), and dual equation if different (currently done).
2. Display commentary:
    - if the equation has commentary display that;
    - elif the equation is equivalent to a lower-numbered equation with commentary, display that;
    - elif the equation is (equivalence + dual) to an equation with commentary, display that with a clear indication that this is the dual equation.

Osalotioman:
This has already been resolved by #1386

Chessing234:
propose #1486

PR #1225, opened 2025-05-18
Commentary updates

blefloch:
Commentary on many equations:
- Merging commentary of dual equations together; I sometimes left it duplicated when the higher-numbered one was much better known (e.g., Littlewood).  This should be revisited depending on what we do in #1218.
- Corrections to commentary that mentioned unknown implications, now all known.
- Those that imply equation 38 or its dual 39, equivalent to sets equipped with some structure.
- Those that imply associativity, whose free group is describable.
- Those that imply commutativity.
- Those that imply idempotence.
- Higher-order ones that were in the tour of selected equations, and some from the literature characterizing groups with a certain exponent, Moufang loops with exponent 3.


teorth:
Fantastic, thanks for all the effort on this.

PR #1241, opened 2025-05-31
fix filename to fix URLs

arnedecadt:
"filename" gave an absolute path in the online version (e.g. click on Equation677_not_implies_Equation255 on https://teorth.github.io/equational_theories/implications/show_proof.html?pair=677,255 ). This should fix it.


teorth:
Thanks!

PR #1247, opened 2025-06-04
[CONTRIBUTIONS] Port Bruno's equation numbering code to ruby.

### LINKED ISSUE #1083: Update scripts to use Bruno's equation numbering code
Follow-up to #482: Many scripts (not `extract_implications` / `export_equations`) currently parse lean source code to get the map of equation id -> equation. Instead, we should use Bruno's algorithm to map equation id -> equation instead, see `scripts/find_equation_id.py`. This allows us to use arbitary equation numbers >4694. Currently there is only a python version in the mentioend script, so it should also be ported to ruby/JS too for those scripts. It would be great to allow the equation explorer to accept/display arbitrary equation ids, even if we don't have implications for them.

Osalotioman:
This pull request:
1.  Ports Bruno's equation numbering code from `find_equation_id.py` to `find_equation_id.rb`, which provides functionality for mapping magma equations to and from their IDs. The script can be used both interactively and with command-line arguments, and it includes a range of mathematical utilities for handling equations.

2. Updates `generate_graphiti_data.rb` to use `find_equation_id.rb` for generating equation list.

Closes #1083

### DIFF
--- a/scripts/find_equation_id.rb
+++ b/scripts/find_equation_id.rb
@@ -84,6 +84,12 @@ class Equation
     def id
       _equation_id(self)
     end
+
+    # Class method
+    def self.from_str(eq_str)
+      # Parse and canonicalize an equation given as a string.
+      _equation_from_str(eq_str)
+    end
     
     def dual
       # Swap all left and right operands, swap lhs and rhs if needed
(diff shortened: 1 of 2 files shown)


vlad902:
Have you tested that this generates correct data matching what was generated before? I am just somewhat surprised that a non-global 'equations' variable is accessible between files.

Osalotioman:
I wrote a script to test it, before updating `generate_graphiti_data.rb`
```
require File.join(__dir__, 'find_equation_id')
equations = {}
for i in 1..4694
    eq = Equation.from_id(i)
    equations[i] = eq.to_s
end
puts equations
```

test_rub

Osalotioman:
> Have you tested that this generates correct data matching what was generated before? I am just somewhat surprised that a non-global 'equations' variable is accessible between files.

It is global because it is defined at the top level of the script, outside of any class or module. In Ruby, top-level class definitions become globally accessible once the file is loaded.

Osalotioman:
I'm getting this error locally when I try to follow the invocation used in `blueprint-paper.yml`. Not sure why it's failing — the same steps seem to work in CI:

```bash
lake exe extract_implications --json --closure --only-implications > /tmp/raw_data/general_implications_closure.json
lake exe extract_implications --json --finite-only --closure --only-implications > /tmp/raw_data/finite_implications_closure.json
lake exe extract_implications unknowns > /tmp/unknowns.json
lake exe extract_implications unknowns --finite-only > /tmp/unknowns.fin.json 
ruby scripts/generate_graphiti_data.rb data/duals.json /tmp/raw_data/general_implications_closure.json /tmp/unknowns.json /tmp/raw_data/finite_implications_closure.json /tmp/unknowns.fin.json > home_page/graphiti/graph.json  
```

Error Message:

```
C:/Ruby34-x64/lib/ruby/3.4.0/json/common.rb:221:in 'JSON::Ext::Parser.parse': unexpected token at '��{' (JSON::ParserError)
        from C:/Ruby34-x64/lib/ruby/3.4.0/json/common.rb:221:in 'JSON.parse'
        from scripts/generate_graphiti_data.rb:41:in 'Object#gen_graph'
        from scripts/generate_graphiti_data.rb:74:in '<main>'
```

The GitHub Actions workflow succeeds [link to run](https://github.com/Osalotioman/equational_theories/actions/runs/15606912759/job/43958475297), so this seems to be something environment-specific.

I’ve traced the issue to `/tmp/raw_data/general_implications_closure.json`, but it’s hard to examine — my editor crashes when opening it.

scripts/find_equation_id.rb
```diff
@@ -0,0 +1,649 @@
+#!/usr/bin/env ruby
+
+=begin
+This module maps magma equations from/to their id
```
vlad902, line 4: Could you just add a note here that this is a port of the python original.

scripts/find_equation_id.rb
```diff
@@ -0,0 +1,649 @@
...
+    ruby find_equation_id.rb -i
+
+or by passing arguments to it from stdin or as arguments: a (space-separated)
+list of ids or of equations (in which the operation can be ".", "*", or "◇"),
+optionally preceeded by "*" to dualize the equation (and characters "[,]" are
+ignored):
+
+    ruby find_equation_id.rb [12, 34] "(w*u)=t*(u*x)" 4567 "*89" "*67" "0=(1*2)*(0*1)"
+
+When used as a module imported in python code, one can use
```
vlad902, line 17: Should this be changed to ruby or deleted? I'm not sure if the API matches the python code here 100%

vlad902:
Thanks for checking that, this PR looks good to me modulo those last two comments.

scripts/find_equation_id.rb
Osalotioman: Done.

Osalotioman:
I ran the workflow again to check for differences.
Everything checks out well as shown below:

IMG_20250615_002009

Link to github actions run:
https://github.com/Osalotioman/equational_theories/actions/runs/15656352156/job/44107905321

vlad902 (approved):

PR #1250, opened 2025-06-13
ci: restore building the paper PDF

Julian:
Refs: #1249

I haven't tested this, hopefully it builds on this pull request to see.
I'm slightly worried it won't immediately work, as nothing in this workflow seems to install latex as a dependency now, but perhaps now that's hidden behind an earlier action being run, which is why I'd like to see if it works or not before suggesting merging.


Julian:
@Vierkantor or @pitmonticone since presumably one of you will immediately know whether this looks right or not.

pitmonticone:
Yes, it was my bad. I'll fix it later. Thanks!

Julian:
I think it's about to fail as it needs to run in the texlive-action image (so I fixed that locally and am just waiting for the failure before pushing). But also happy to just leave it to the experts on the CI setup!

PR #1253, opened 2025-06-16
ci: Bump update action and set to stable releases only.

Vierkantor:
This new version does not error if no upgrades exist, and has a 'stable' option for skipping prereleases and Mathlib `master` commits. We enable this option, reflecting the preferences of this project.

Tested on https://github.com/Vierkantor/lean-update-tester.


pitmonticone (approved):
Thanks!

PR #1256, opened 2025-06-24
Commentary for June

blefloch:
- Add many names appearing in somewhat obscure literature.
- Counting magmas for some low-order laws (I'll write up something much more detailed for the blueprint).
- Equivalent characterizations for some laws.
- Some free magmas.


teorth:
Thanks for this!  This is making Equation Explorer a valuable resource for the broader field beyond the original ETP goals.

There is some weird CI problem that I don't understand, but once that is resolved, I will merge this PR.

PR #1258, opened 2025-06-24
refactor: Apply find_equation_id.rb (ported from find_equation_id.py) to related files

### LINKED ISSUE #1083: Update scripts to use Bruno's equation numbering code

Osalotioman:
`find_equation_id.rb` is used in:
- `equational_theories/Generated/EquationSearch/src/search.rb` to generate a bijective map from equation to id and back.
- `scripts/generate_graphviz_graph.rb` to generate equation list.

Closes #1083

### DIFF
--- a/equational_theories/Generated/EquationSearch/src/search.rb
+++ b/equational_theories/Generated/EquationSearch/src/search.rb
@@ -36,6 +36,7 @@ SEARCH_MAX_EXPR_OPERATIONS_ALLOWED = 5
 
 
 require File.join(__dir__, '../../../../scripts/graph')
+require File.join(__dir__, '../../../../scripts/find_equation_id')
 
 STRICT_ERROR_CHECKING=true
 VERBOSE = false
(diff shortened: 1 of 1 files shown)


Osalotioman:
awaiting-review

vlad902:
As per [#Equational > Update scripts to use Bruno's equation numbering code. @ 💬](https://leanprover.zulipchat.com/#narrow/channel/458659-Equational/topic/Update.20scripts.20to.20use.20Bruno's.20equation.20numbering.20code.2E/near/525659599) I'd suggest closing instead of spending the time to check that these vestigial unused scripts are functioning correctly.

PR #1264, opened 2025-07-12
Minor fixes for typos etc.

dsm054:
Added some missing equations, fixed some typos, and standardized some spellings.  Left things we should probably standardize (e.g. non-commutative or noncommutative) alone.


teorth:
I think it's fine to make an arbitrary choice for standardization for e.g., noncommutative vs. non-commutative and jut go ahead and implement it globally; making it consistent is more important than any minor preferences on way or the other, in my opinion.

PR #1289, opened 2025-09-17
Section of the paper on the full spectrum problem (closes #1029)

### LINKED ISSUE #1029: WRITING: write a section of the paper on the full spectrum problem
The results established in https://leanprover.zulipchat.com/#narrow/channel/458659-Equational/topic/Equations.20with.20full.20spectrum can be written up in the paper, or at least summarized with fuller details in the blueprint.

blefloch:
For laws up to order 4 Vlad Tsyrklevich has formalized the proofs so we're safe.  I only mention in passing the work on higher order laws, which requires more complicated constructions and probably ought to involve cleaning up some of my Python scripts.


teorth:
Thanks!

PR #1338, opened 2025-10-25
change obey to satisfy

teorth:


teorth:
This was following the suggestion in https://leanprover.zulipchat.com/#narrow/channel/458659-Equational/topic/Writing.20the.20paper/near/547041617 to harmonize the notation in the paper (and, by extension, the rest of the project).

PR #1369, opened 2025-11-04
Forgot to add myself to author block

carlini:
I messed up and only added myself to two of the three places I should have.

(I noticed this when looking at the author block, it looks like in #1368 @amirlb may have accidentally inserted in the wrong position at the end instead of alphabetically. I have not fixed that problem because I'm not sure if that was intentional, but noting that here in case it should be corrected.)


amirlb:
It was entirely accidentally, I didn't mean anything by it.

Shall I open a new PR to fix this?

Issue #848, opened 2024-11-18
EQUATION EXPLORER: Allow searching/displaying any equation


vlad902:
With Bruno's new fast algorithm for mapping equation to id and back, there's no reason the equation explorer can't handle arbitrary equation numbers or formulas, even if we don't have implications it could still be useful as a web-based find_equation_id.

teorth:
This can be combined with #378.

vlad902:
It could be, though they're really two separate tasks.

blefloch:
Note that my python code is efficient in almost all cases, with the exception of equations with the same shape on the left-hand side and right-hand side: then it has to list all equations in order (basically).  This could probably be solved if there is appetite for it, but the combinatorics of how to count equations up to isomorphism in that case were a bit involved.

Osalotioman:
claim

Osalotioman:
@teorth this issue can be closed now, as it was resolved when you merged #1377 
Since #1377 contained the updates from #1376

PR #1376, opened 2025-11-05
[FEATURE] Allow displaying of higher order equations by explorer

Osalotioman:
Before addressing this pr, you might want to take a look at #1375 
This pr was broken up from #1288
This pr resolves issue #848

- Port `scripts/find_equation_id.py` to js `scripts/find_equation_id.js`
- Copy `scripts/find_equation_id.js` and modify for use in the equation explorer as `home_page/implications/find_equation_id_1.js` suffixing with `_1` since  `home_page/implications/find_equation_id.js` already exists but does not seem to be a proper port of `scripts/find_equation_id.py`
- Use `home_page/implications/find_equation_id_1.js` to display equations of higher order along with its dual.
- Preserve code no longer in use from `home_page/implications/find_equation_id.js` as per discussion [here](https://leanprover.zulipchat.com/#narrow/channel/458659-Equational/topic/Update.20scripts.20to.20use.20Bruno's.20equation.20numbering.20code.2E)

[Link to thread on Zulip](https://leanprover.zulipchat.com/#narrow/channel/458659-Equational/topic/.23848/with/553945753)

### DIFF
--- a/home_page/implications/script.js
+++ b/home_page/implications/script.js
@@ -346,6 +346,15 @@ function renderImplications(index) {
     let dualIndex = dualEq.id;
     if (dualIndex !== null) {
 	    selectedEquationDual.innerHTML = "(Dual equation: <a class='link' onclick='renderImplications("+(dualIndex-1)+");'>" + `Equation${dualIndex}[${dualEq}]` + "</a>)";
+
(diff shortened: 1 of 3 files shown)


Osalotioman:
There is no need to revert. I have resolve the conflicts so it can be merged cleanly.

PR #1379, opened 2025-11-06
[FIX] Add target attribute to links for opening in new tab

Osalotioman:
Clicking on the link just change the url instead of taking one to the new page, this update makes the browser open up the link in a new tab so it should work fine.

[Link to thread on Zulip](https://leanprover.zulipchat.com/#narrow/channel/458659-Equational/topic/.23848/with/553945753)

### DIFF
--- a/tools/fme/dist/elm.js
+++ b/tools/fme/dist/elm.js
@@ -5520,11 +5520,17 @@ var $author$project$Main$update = F2(
 					var urlRequest = msg.a;
 					if (!urlRequest.$) {
 						var url = urlRequest.a;
-						return _Utils_Tuple2(
+						return A2(
+							$elm$core$String$startsWith,
+							'/equational_theories/fme/',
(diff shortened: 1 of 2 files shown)


Osalotioman:
This is what happens if the links are left like that, it works locally, but on the main site, it just changes the urls.

https://github.com/user-attachments/assets/59ef58a0-402d-4a70-98b9-c22cbb04af0d


Locally when the urls are used directly:

https://github.com/user-attachments/assets/b460d5b9-afac-42f5-bd4e-4d9025f9a1ff

Osalotioman:
> I'm not sure, why are we using a new tab here? None of the other interfaces use this pattern.

You created a pr to do the same earlier in the project: https://github.com/teorth/equational_theories/pull/808

Osalotioman:
I have added this change, and updated the url not to open in a new tab.

vlad902 (approved):

PR #1383, opened 2025-11-07
[FEATURE] Allow displaying of higher order equations by explorer

Osalotioman:
This PR was remade after reversion of #1376 so due review process can be followed.

### DIFF
--- a/home_page/implications/script.js
+++ b/home_page/implications/script.js
@@ -346,13 +346,7 @@ function renderImplications(index) {
         selectedEquationDual.innerHTML = "";
     }
 
-    const equationsArray = (typeof equations !== 'undefined' && Array.isArray(equations)) ? equations : null;
-    const commentaryArray = (typeof commentary !== 'undefined' && Array.isArray(commentary)) ? commentary : null;
-    const totalEquations = equationsArray ? BigInt(equationsArray.length) : null;
-    const maxAddressableIndex = BigInt(Number.MAX_SAFE_INTEGER);
-    const datasetAvailable = totalEquations !== null && bigIndex < totalEquations && bigIndex <= maxAddressableIndex;
-
-    if (!datasetAvailable) {
+    if (!(bigIndex < BigInt(equations.length))) {
         hideVisibility("equationCommentary");
         equationCommentary.innerHTML = "";
 
@@ -391,9 +385,9 @@ function renderImplications(index) {
     showVisibility("smallestMagmaLink");
 
     const commentaryIndex = indexNumber + 1;
-    if (commentaryArray && commentaryArray[commentaryIndex] !== undefined) {
+    if (commentary && commentary[commentaryIndex] !== undefined) {
         showVisibility("equationCommentary");
-        equationCommentary.innerHTML = commentaryArray[commentaryIndex];
+        equationCommentary.innerHTML = commentary[commentaryIndex];
     } else {
         hideVisibility("equationCommentary");
         equationCommentary.innerHTML = "";


home_page/implications/find_equation_id.js
```diff
@@ -18,12 +18,12 @@ function closePopup(element_id) {
     document.getElementById(element_id).style.display = 'none';
 }
 
-function tokenize(expr) {
-    return expr.replace(/\./g, '◇').replace(/\*/g, '◇').replace(/\(/g, ' ( ').replace(/\)/g, ' ) ').replace(/◇/g, ' ◇ ')
-        .split(' ')
-        .filter(token => token.trim() !== '');
-}
-
+//function tokenize(expr) {
```
vlad902: Ditto, delete this and the functions below also seem to be unused now so also delete them.

home_page/implications/find_equation_id.js
```diff
@@ -239,13 +1028,14 @@ function findEquation() {
     try {
         // Determine if user entered integer input or equation input
         if (!isNaN(inputEq) && (/^\d+$/.test(inputEq))){
-            eqNum = parseInt(inputEq, 10) - 1;
+            eqNum = Equation.fromId(parseInt(inputEq)).id;
```
vlad902: This seems unnecessary, it's calculating the equational form but then throwing away all the data to get back the equation number. We should keep the old parseInt.

Osalotioman: I used `Number` instead of `parseInt` cause `parseInt` fails for input like ` 8 2`, it spits out `8`.

home_page/implications/find_equation_id.js
```diff
@@ -18,219 +11,794 @@ function closePopup(element_id) {
...
-            return tokens.shift();
+            throw new Error("Unexpected end of expression");
         }
-        if (tokens[0] === '(') {
-            tokens.shift();
+        if (tokens[0] === "(") {
+            tokens.shift(); // Remove opening parenthesis
             const left = parseElement();
-            if (tokens.length === 0 || tokens[0] !== '◇') {
-                throw new Error(`Expected '◇' after '${left}' in parentheses. Did you mean to write '(${left} ◇ ...)'?`);
```
vlad902: The previous error messages seem more helpful, lets keep them here and below.

home_page/implications/find_equation_id.js
```diff
@@ -239,17 +807,17 @@ function findEquation() {
     try {
         // Determine if user entered integer input or equation input
         if (!isNaN(inputEq) && (/^\d+$/.test(inputEq))){
-            eqNum = parseInt(inputEq, 10) - 1;
-        }
-        else{
-            eqNum = findEquationNumber(inputEq)-1;
+            eqNum = Number(inputEq);
```
vlad902: Now that we support large numbers this can overflow or lose precision past `2**53` because of JS representing ints using float. Do you know if everything works continues to work if we instead do `eqNum = BigInt(inputEq)` here? If so, it'd be nice to use BigInt here.

Also from a quick test it seems the code above is also vulnerable to returning the wrong number if given an equation in this high range, e.g.
``` x = y ◇ (((((z ◇ w) ◇ ((y ◇ z) ◇ (u ◇ v))) ◇ (w ◇ (z ◇ ((y ◇ w) ◇ y)))) ◇ (v ◇ r)) ◇ v)```
is equation 9007199254740995 but find_equation_id.js will say its 9007199254740996 due to numerical imprecision. Would it be hard to change the above code to use BigInt throughout so we have infinite precision?

File: home_page/implications/find_equation_id.js (hunk 3 above)
Comment (Osalotioman) [reply home_page/implications/find_equation_id.js]: Done, I've brought back the previous error messages which were more helpful.

File: home_page/implications/find_equation_id.js (hunk 4 above)
Comment (Osalotioman) [reply home_page/implications/find_equation_id.js]: Everything does not continue to work if we do `eqNum = BigInt(inputEq)` here, we'll need to use big int through out for it to work.

Using bigint throughout this file worked after numerous iterations:
image

but then new errors pop up in `home_page/implications/script.js` which I am currently trying to tackle:
image

Comment (Osalotioman) [reply home_page/implications/find_equation_id.js]: I've pushed the updates.

File: home_page/implications/script.js [hunk 5]
```diff
@@ -311,56 +311,104 @@ function renderEquationList(sortBy = 'index', sortOrder = 'asc') {
...
+    if (dualIndex !== null) {
+        const dualDisplay = (dualIndex - 1n).toString();
+        selectedEquationDual.innerHTML = `(Dual equation: <a class='link' onclick="renderImplications('${dualDisplay}')">Equation${dualIndex.toString()}[${dualEq}]</a>)`;
     } else {
-        showVisibility("equationCommentary")
-        equationCommentary.innerHTML = commentary[index+1];
+        selectedEquationDual.innerHTML = "";
     }
 
+    const equationsArray = (typeof equations !== 'undefined' && Array.isArray(equations)) ? equations : null;
```
vlad902: Not sure why we need this for equations/commentary, if they're not defined it seems fine to just crash. Unless I'm missing some case when the explorer runs and they're actually legitimately not defined?

home_page/implications/script.js
```diff
@@ -311,56 +311,104 @@ function renderEquationList(sortBy = 'index', sortOrder = 'asc') {
...
+    const commentaryArray = (typeof commentary !== 'undefined' && Array.isArray(commentary)) ? commentary : null;
+    const totalEquations = equationsArray ? BigInt(equationsArray.length) : null;
+    const maxAddressableIndex = BigInt(Number.MAX_SAFE_INTEGER);
```
vlad902: Ditto, I think just checking against `equations.length` is fine.

vlad902 (commented):
Some minor comments but looks good otherwise.

PR #1386, opened 2025-11-25
[ENHANCE] Improve Commentary Display

Osalotioman:
This PR is a remake of #1384 to make review easier after merging #1383

### DIFF
--- a/home_page/implications/script.js
+++ b/home_page/implications/script.js
@@ -403,7 +403,7 @@ function renderImplications(index) {
     document.getElementById('equivalentEquations').innerHTML = equivalentEquationsHtml;
 
     const baseEquivalentEquationId = equivalentClass[0];
-    const dualEquivalentClass = equiv.find(cls => cls.includes(dualIndex - 1n)) || [dualIndex - 1n];
+    const dualEquivalentClass = equiv.find(cls => cls.includes(Number(dualIndex - 1n))) || [Number(dualIndex - 1n)];
     const baseDualEquivalentEquationId = dualEquivalentClass[0];
     if (commentary[eqId] === undefined && commentary[dualIndex] === undefined && commentary[baseEquivalentEquationId + 1] !== undefined) {
         showVisibility("equationCommentary");


home_page/implications/script.js
```diff
@@ -333,23 +333,46 @@ function renderImplications(index) {
     const eq = Equation.fromId(eqId);
     updateUrl(eqId);
 
+    let equationHasCommentary = false;
```
vlad902: I don't like introducing these new variables that make it more difficult to track what's going on and what means what. Just checking `if (commentary[bigIndex])` is easier to understand.

home_page/implications/script.js
```diff
@@ -333,23 +333,46 @@ function renderImplications(index) {
...
     selectedEquation.dataset.index = bigIndex.toString();
 
+    // Check if equation can be converted to Number safely
+    let indexNumber = null;
+
+    if (bigIndex <= BigInt(Number.MAX_SAFE_INTEGER)) {
+        indexNumber = Number(bigIndex);
+
+        const commentaryIndex = indexNumber + 1;
+        if (commentary && commentary[commentaryIndex] !== undefined) {
```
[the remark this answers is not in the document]
Osalotioman: Check for commentary has been removed.

home_page/implications/script.js
```diff
@@ -337,19 +337,29 @@ function renderImplications(index) {
     selectedEquation.textContent = `Equation${eqId.toString()}[${eq.toString()}]`;
     selectedEquation.dataset.index = bigIndex.toString();
 
+    if (commentary[eqId] !== undefined) {
```
vlad902: It seems more natural to me to move this if() to the bottom, then we can have a long series if() else if() else if()s about the commentary all in one place instead of scattered around.

home_page/implications/script.js
```diff
@@ -337,19 +337,29 @@ function renderImplications(index) {
...
+        equationCommentary.innerHTML = "";
+    }
+
     const dualEq = eq.dual();
     const dualIndex = dualEq.id;
     if (dualIndex !== null) {
         const dualDisplay = (dualIndex - 1n).toString();
         selectedEquationDual.innerHTML = `(Dual equation: <a class='link' onclick="renderImplications('${dualDisplay}')">Equation${dualIndex.toString()}[${dualEq}]</a>)`;
+
+        if(commentary[eqId] === undefined && commentary[dualIndex] !== undefined) {
```
vlad902: this should be part of that big if() else if() block

home_page/implications/script.js
```diff
@@ -406,6 +407,31 @@ function renderImplications(index) {
...
+    if (commentary[eqId] === undefined && commentary[dualIndex] === undefined && commentary[baseEquivalentEquationId + 1] !== undefined) {
+        showVisibility("equationCommentary");
+        equationCommentary.innerHTML = `
+            <h2>Commentary of the equivalent ${equations[baseEquivalentEquationId]}:</h2>
+            ${commentary[baseEquivalentEquationId + 1]}
+        `;
+    }
+
+    // Try converting the dual index to Number safely
+    let dualIndexNumber = null;
```
vlad902: We only have implications for equations <4694 at the moment and if we ever add support for higher-numbered equations AND then add any equation above 2**53 we would have to first refactor to the explorer to support it so I think it is safe to not check for this here.

Osalotioman: I've taken out the check.

home_page/implications/script.js
Osalotioman: If I move this block under this check `if (bigIndex > BigInt(equations.length - 1)) {` to meet up with the other commentary checks relating to equivalent equations, then equations with id greater than 4694 such as `Equation599055[x = ((((y ◇ z) ◇ (x ◇ z)) ◇ z) ◇ y) ◇ y]` which have commentary, will not show any.

Currently, I've grouped the commentary checks above and below this check `if (bigIndex > BigInt(equations.length - 1)) {` into if else blocks.

I'm thinking of the cleanest way to bring them together currently, without altering too many things while preserving functionality.

home_page/implications/script.js
```diff
@@ -339,17 +343,18 @@ function renderImplications(index) {
...
+    const dualDisplay = (dualIndex - 1n).toString();
+    selectedEquationDual.innerHTML = `(Dual equation: <a class='link' onclick="renderImplications('${dualDisplay}')">Equation${dualIndex.toString()}[${dualEq}]</a>)`;
 
-    if (!(bigIndex < BigInt(equations.length))) {
-        hideVisibility("equationCommentary");
-        equationCommentary.innerHTML = "";
+    if (commentary[eqId] !== undefined) {
+        showVisibility("equationCommentary");
+        equationCommentary.innerHTML = commentary[eqId];
+    }else if(commentary[dualIndex] !== undefined) {
```
vlad902:
```suggestion
    } else if(commentary[dualIndex] !== undefined) {
```

Osalotioman:
Don't merge yet, I just noticed an issue with 1090.

Osalotioman:
Fixed.

This was the issue: https://github.com/teorth/equational_theories/pull/1386#discussion_r2580299248

PR #1427, opened 2026-01-31
Correct find_equation_id Python script for tautological equations

blefloch:
- Return an id of 0 for tautological equations other than Equation 1
- Avoid wrong Equation.id results when the equation is non-canonical


Chessing234:
Same tautology id bug was still in the JS ; opened a small follow-up.

PR #1437, opened 2026-05-06
chore: Bump to version 4.29

Timeroot:
This was a long and messy bump. I also did a bunch of incidental cleanup (in the form of resolving lints) along the way, which was honestly necessary to be able to find the fixes and so on, pretty often.


Timeroot:
Looks ready!

PR #1443, opened 2026-06-01
fix: emit dual of each proven implication in extract_implications

### LINKED ISSUE #1245: In the finite case, the extract_implications script omits some dual implications
In the finite case, the dual laws 1113 and 2534 are equivalent (but not in general).  So I would expect `lake exe extract_implications --finite-only` to include both implications 1113→2534 and 2534→1113.  But the second one does not appear.  In fact, the only implications (in the output of that script) starting with 2534 are 2534→2441 and 2441→1.

1. In the top directory, run `lake build`
2. Run
    ```
    $ lake exe extract_implications --finite-only | grep '^Equation1113'
    Equation1113 → Equation1020
    Equation1113 → Equation2534
    ```
3. If you want, compare to the result without the `--finite-only` flag: it does not have the second line.
4. Run
    ```
    $ lake exe extract_implications --finite-only | grep '^Equation2534'
    Equation2534 → Equation2441
    $ lake exe extract_implications --finite-only | grep '^Equation2441'
    Equation2441 → Equation1
    ```
5. Note that there is a single implication (to the dual of 1020), and that this law 2441 only implies 1, so the problem is not simply fixed by taking some transitive closure.  The script correctly tells us that 1113→2534 but I would expect it to tell us also that 2534→1113.

Possibly the resolution is that I have to take into account non-implications when computing the transitive closure?

Chessing234:
## Summary

When `lake exe extract_implications` lists proven implications (without `--closure`), it only printed edges extracted directly from Lean theorems. If `Equation1113 → Equation2534` is proven and 1113/2534 are dual laws, the dual implication `Equation2534 → Equation1113` was missing even though it follows by magma duality.

## Root cause

`generateOutput` called `Closure.toEdges` alone in the non-closure path. Dualization is wired into `closure_aux`, but that path is only used with `--closure`.

## Why this fix is correct

For each direct implication edge, we also append its dual (via `DualityRelation.dual`). Proving `A → B` yields the dual proof `dual(A) → dual(B)`; when `A` and `B` are dual pairs this is the reverse implication the reporter expected under `--finite-only`.

Fixes #1245

## Test plan
- [ ] `lake build`
- [ ] `lake exe extract_implications --finite-only | grep '^Equation1113'` shows both directions when the finite implication is proven

### DIFF
--- a/scripts/extract_implications.lean
+++ b/scripts/extract_implications.lean
@@ -154,7 +154,12 @@ def generateOutput (inp : Cli.Parsed) : IO UInt32 := do
     let rs := matchFinite rs finite_only
     let rs := if only_implications then rs.filter (·.variant matches .implication ..) else rs
     let rs := rs.map (·.variant)
-    let rs ← if include_impl then Closure.closure rs dualityRelation.dualEquations else pure (Closure.toEdges rs)
+    let rs ← if include_impl then Closure.closure rs dualityRelation.dualEquations else do
+      let edges := Closure.toEdges rs
+      let duals := edges.filterMap fun
+        | .implication imp => (dualityRelation.dual imp).map (.implication ·)
+        | _ => none
+      pure (edges ++ duals)
     if inp.hasFlag "json" then
       let implications := (rs.filter (·.isTrue)).map (·.get)
       let nonimplications := (rs.filter (!·.isTrue)).map (·.get)


Chessing234:
awaiting-review

Chessing234:
Rebased onto latest main and pushed to retrigger CI.

Chessing234:
Pushed an empty commit to retrigger CI on latest main.

Chessing234:
Latest rebase build passed; this should be ready for review.

Chessing234:
awaiting-review

Issue #1411, opened 2025-12-09
Tautological equations are parsed incorrectly by the Equation Explorer


blefloch:
**Describe the bug**
On the Equation Explorer entering a tautological equation other than `x=x` results in displaying the equation that would follow that tautological equation.

**To Reproduce**
1. Go to https://teorth.github.io/equational_theories/implications/
2. In the first text box, type `x*(y*x)=x*(y*x)`
3. Click `Jump to Equation`
4. Observe that the equation displayed is the non-trivial [Equation 4314](https://teorth.github.io/equational_theories/implications/?4314) `x◇(y◇x) = x◇(y◇y)`.

**Expected behavior**
Either a message stating that the equation is tautological, or silently pointing to https://teorth.github.io/equational_theories/implications/?1 would be better than showing an equation that is different from the one that was input.

blefloch:
A popup is a good idea, but it should just point to Equation 1 I think, since Equation 4314 is mathematically unrelated to the one requested; there are countless one-character changes leading to valid equations and that particular one has not particular relationship to the requested one `x◇(y◇x) = x◇(y◇x)`.

PR #1450, opened 2026-06-23
fix: return id 0 for tautological equations in JS canonicalizer

Chessing234:
## Summary
- `_equationId` in `find_equation_id.js` could assign a nonzero id to tautological equations such as `x*(y*x)=x*(y*x)`.

## Root cause
#1427 fixed this in `find_equation_id.py` by returning `0` when the rhyme equals its lhs/rhs rotation, but the JS port still counted such rhymes in the slow balanced branch.

## Fix
Mirror the Python early return before the rhyme enumeration loop.

## Test plan
- [ ] `Equation.fromStr('x*(y*x)=x*(y*x)').id === 0n` in the browser helper

### DIFF
--- a/home_page/implications/find_equation_id.js
+++ b/home_page/implications/find_equation_id.js
@@ -599,6 +599,11 @@ function _equationId(inputEq) {
     if (l !== r) {
         pid = findRhymeId(inputEq.rhyme);
     } else {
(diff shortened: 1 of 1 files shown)


Osalotioman:
See #1411

PR #1453, opened 2026-06-28
fix: add permalink to full spectrum Zulip thread

Chessing234:
## Summary
- Pin the FullSpectrum header link to the discussion message, matching commentary pages.

## Test plan
- [ ] Link opens the intended Zulip thread


Chessing234:
Topic-only link was easy to lose in a busy stream — pinned it like the commentary pages.

PR #1454, opened 2026-06-28
fix: add Zulip permalinks for twin-pairs commentary links

Chessing234:
## Summary
- Add `/near/` permalinks to the twin-pairs and 476~503 Zulip links in commentary for equations 124, 476, 503, and 1119
- Switch the numerical-coincidence links to the channel URL format while adding anchors

## Test plan
- [ ] Click each updated link and confirm it opens the intended Zulip message


Chessing234:
The twin-pairs links pointed at whole topics with no message anchor, so they landed at the top of long threads. Added /near/ IDs for the opening posts.

PR #1455, opened 2026-06-28
fix: add Zulip permalinks for Obelix and Littlewood commentary

Chessing234:
## Summary
- Add `/near/` permalinks to the Obelix construction link in `Equation1491.md`.
- Add `/near/` permalink to the Outstanding equations thread link in `Equation1086.md`.

## Test plan
- [ ] Click both commentary links and confirm they land on the intended Zulip messages.


Chessing234:
These two commentary links only pointed at the Zulip topic, not the specific message — added /near/ IDs so they jump to Pace's Obelix construction post and Tao's Outstanding equations summary.

PR #1456, opened 2026-06-28
fix: add Zulip permalinks for eq 124/1648 commentary threads

Chessing234:
## Summary
- Add `/near/` permalinks to the 1076 and 1648 implication-thread links in `Equation124.md` and `Equation1648.md`.

## Test plan
- [ ] Click both commentary links and confirm they land on the intended opening Zulip messages.


Chessing234:
The 124/1648 commentary had two Zulip links with no /near/ ID, so they opened at the top of the thread instead of the posts being cited.

PR #1476, opened 2026-08-29
Load Eqns*.lean when running process_implications CLI

Chessing234:
## Summary
- Standalone `process_implications.py <file.lean>` passed an empty equations list, so the universe stayed empty.
- Discover `equational_theories/Equations/Eqns*.lean` by default (same sources `All.lean` imports).

## Test plan
- [ ] `python scripts/process_implications.py equational_theories/Subgraph.lean` reports a non-empty universe / unknown count

### DIFF
--- a/scripts/process_implications.py
+++ b/scripts/process_implications.py
@@ -14,6 +14,7 @@ import os
 from random import sample
 import re
 from sys import argv, stdin
+from pathlib import Path
 import networkx as nx
 
 
@@ -71,6 +72,18 @@ def get_unknown_implications(universe, known_implies, known_not_implies):
         - all_negative_implications
     )
 
+
+
+def default_equations_files():
+    """Equation definition files used to populate the universe for CLI runs."""
+    root = Path(__file__).resolve().parent.parent
+    eq_dir = root / "equational_theories" / "Equations"
+    files = sorted(eq_dir.glob("Eqns*.lean"))
+    if not files:
+        raise FileNotFoundError(
+            f"No Eqns*.lean files found under {eq_dir}; pass equation files explicitly."
+        )
+    return [str(f) for f in files]
 
 def parse_proofs_file_internal(
     universe, known_implies, known_not_implies, equations_files, file_name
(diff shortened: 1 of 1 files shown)


teorth:
Thanks for this, but I don't think we should take it — and the reason is in your own #1468, which is now merged.

**The premise doesn't hold.** The summary says "passed an empty equations list, so the universe stayed empty". The equations list is empty, but the universe is not: `parse_proofs_file_internal` populates it from the theorems it parses out of the target `.lean` file. That is why `python scripts/process_implications.py equational_theories/Subgraph.lean` on `main` reports 862 unknown implications over a universe of 43 equations, not zero. The test plan's success criterion — "reports a non-empty universe / unknown count" — is already met before this change.

**Measured effect on `Subgraph.lean`:**

| | main | this PR |
|---|---|---|
| runtime | 0.5 s | 10.3 s |
| peak RSS | — | **3.6 GB** |
| reported unknowns | 862 | **21,948,323** |

The 21.9M figure is the point. With the full `Eqns*.lean` universe loaded, `get_unknown_implications` materialises `universe × universe`, and since `Subgraph.lean` mentions only 43 of the 4694 laws, essentially every pair is "unknown". The tool then prints a random sample of 10 from those 21.9 million. That is strictly less useful than the 862 the CLI reports today, which are the implications this particular file genuinely leaves open.

**You already worked this out.** From the body of #1468, which is now the commit message on `main`:

> I deliberately did **not** start feeding all 4694 equations into the CLI path. `get_unknown_implications` materialises `universe × universe`, so that would go from ~1.8k pairs to ~22M.

That prediction was accurate to within a rounding error.

If there is a real use case behind this — say, wanting to ask which of the full 4694 a given file leaves open — the fix is not to enlarge the universe but to change what `get_unknown_implications` enumerates, so it does not build the full square. A `--equations` flag defaulting to off would also be a much safer shape than changing the default for every caller, and would match `generate_image.py`, which already takes equation files as an explicit option.

Closing for now; happy to look again if you want to take one of those routes.

PR #1492, opened 2026-09-07
paper: record the reduced role of the blueprint in the ETP

teorth:
Follow-up to #1491.

## The change

`intro.tex` splits the logistics paragraph and adds the difference between the ETP and its predecessors that Referee #1 asked about in R1.12:

> One key difference worth noting, however, was the reduced role of the Lean blueprint, which was used to organize the formalization of a few selected implications, and to record various observations made informally, but was not the centralized source of project objectives as in previous projects; instead, the project relied both on informal Zulip chats and on the formal issue tracking system in Github.

The mention of the GitHub repository earlier in the section is also sharpened to name its issue tracking and pull request system specifically.

## Why this matters for the response

Referee #1 offered two ways to answer R1.12: discuss how the ETP differs from the Liquid Tensor Experiment, or accept that "the more accurate framing could be that the ETP experimentally confirmed the model used in the LTE". In #1491 we took the second. This change supplies the first, and the two now sit together — the ETP confirms the general paradigm while differing on where the project's objectives lived.

`response.tex` is updated for R1.12 accordingly, since it previously understated what the paper says. No change to the item's status, which was already done.

## Verification

`main.tex` builds at 76 pages with 0 errors and 0 undefined references or citations. `response.tex` builds clean at 8 pages.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Shreyas4991:
Is the AI OSS slop citation already in the references file?

teorth:
> Is the AI OSS slop citation already in the references file?

It was added in the previous PR. https://github.com/teorth/equational_theories/pull/1491/changes#diff-94819accac1b115234cd9ed6df1e3b31d894c6d485c30c2189f6ce5f78c5ffa0R1308

paper/intro.tex
```diff
@@ -40,7 +40,7 @@ \subsection{The Equational Theories Project}
...
...mial--Freiman--Ruzsa conjecture~\cite{PFR_Tao_Dilles_2023}, and the infrastructure developed for such projects, such as the Lean blueprint software~\cite{leanblueprint} and the Lean Zulip channel~\cite{zulip_lean}, and the integration with a Github repository (and its issue tracking and pull request system in particular), which in particular allowed for a more decentralized approach than with older formalization projects such as Flyspeck~\cite{flyspeck}, the odd order theorem formalization~\cite{odd_order}, AFP~\cite{afp}, or math-comp~\cite{mathcomp}.  However, as this project required relatively little advanced mathematics, it is likely that one could have achieved broadly similar results using other standard formal proof assistants that have not developed as extensive a library of mathematics as Lean's \emph{Mathlib}. For instance, \emph{Isabelle}~\cite{isabelle_hol, sledgehammer_smt} would be a particularly plausible alternative proof assistant for this project, due to its native support of SMT solvers via the \emph{Sledgehammer} tool; in contrast, as discussed in \Cref{automated-sec}, custom conversions from ATP-generated proofs to Lean code had to be written for this project.
```
Shreyas4991: Maybe it is valuable to point out that we didn't quite have an idea of what math we might require in the beginning, especially for the counterexamples. So starting with mathlib seemed prudent at the time. 

Further, a lot of lean tactics start out in mathlib and don't exist in core lean. We had more freedom to use them simply because we used mathlib. Maybe we need to highlight one or two such instances.

teorth: Good point; I have edited.

PR #1496, opened 2026-09-08
Remove the raw referee reports from the repository

teorth:
The editors have asked that the referee reports not be public.

## What this removes

- `paper/referee reports/report 2.txt` — Referee #2's report, which was tracked and pushed.
- `paper/referee reports/response-todo-zulip.md` — the Markdown to-do list, which paraphrased **both** reports request by request.

`paper/referee reports/report 1.pdf` was never committed (the blanket `*.pdf` rule caught it), and has been taken out of the working tree as well.

## What stays

`paper/referee reports/response.tex` and its built PDF, which the editors are content to keep public.

## Guard against recurrence

A `.gitignore` rule now covers `paper/referee reports/report*`. Verified: both report files are ignored, while `response.tex` remains tracked and unignored.

## Please note — this does not erase the public history

Both removed files were pushed previously, so their contents are still reachable through `git log` / `git show` on earlier commits, and through the GitHub UI. Removing them from the working tree is not the same as removing them from the repository's history.

If the editors' request is meant to cover the history too, that needs a rewrite (`git filter-repo` or equivalent) and a force push to `main`, and even then GitHub may retain the old blobs until asked to garbage-collect them. That is a destructive operation on a public default branch with outside contributors, so it should be a deliberate decision rather than part of this PR.

Local copies of both reports have been kept outside the repository.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Osalotioman (commented):
Would the commit history not have to be rewritten?

PR #1499, opened 2026-09-09
Patch vulnerable widget dependencies and regenerate the bundle

Chessing234:
Refresh the widget lockfile to patched compatible versions, upgrade @rollup/plugin-terser to 1.0.0 to remove its vulnerable serialize-javascript dependency, and record its Node >=20 requirement. Regenerate the tracked HasseDiagram.js bundle so the checked-in output reflects the dependency changes.

This addresses published advisories including [Rollup path traversal](https://github.com/advisories/GHSA-mw96-cpmx-2vgc), [serialize-javascript code execution](https://github.com/advisories/GHSA-5c6j-r48x-rmvq), and lodash/Immutable/glob dependency advisories. npm audit drops from 17 affected dependency entries (9 high) to two moderate entries, with no high findings remaining.

Validation: clean npm ci and npm run build (TypeScript and production Rollup) pass. The remaining entries are react-d3-tree and its uuid dependency, which has no compatible upstream fix. The tree library calls only uuid.v4(), whereas [the advisory](https://github.com/advisories/GHSA-w5hq-g745-h8pq) concerns caller-provided buffers in v3/v5/v6. No override or audit suppression is added. A live Lean infoview session was not exercised.

Prepared with Codex assistance.


vlad902:
I'm not sure this is valuable, this widget as far as I can tell was never used and should just be removed.

Chessing234:
agreed — nothing imports Visualization.lean and the widget isn't wired into the build path. closing the dep patch in favor of removal; i'll open a separate pr to drop widget/ and Visualization.lean unless you'd rather handle it.

PR #1500, opened 2026-09-09
paper(R2.T10): replace/remove spartan for clarity

Osalotioman:


Shreyas4991:
~~How does removing Spartan address the concern R2.10? I put this word with some thought and I don't see  reviewer 2 objecting to it in point 10~~

Oh I see point r2.T10. In this case I would suggest calling it a very small kernel (8000 lines). 

EDIT: the sentence is meant to suggest that the kernel is small and simple and thus more trusted.

paper/project.tex
```diff
@@ -62,7 +62,7 @@ \subsection{Organizing the collaboration in this project}
...
...f mathematical foundations. \emph{Lean} 4 has the calculus of constructions extended by inductive types as its core logic. This logic is sufficiently powerful to express mathematical definitions and theorems for almost all areas of mathematical interest, while being relatively minimal and easy to write proof checkers for. Additionally, modern ITPs provide a convenient programming language which helps express mathematical ideas in a syntax closer to a mathematician's intuition than would be permitted by raw logical terms. A subset of ITPs like \emph{Lean}, \emph{Rocq}, and \emph{Isabelle} go one step further and provide the means to generate proofs through so-called \emph{tactics}. There are usually numerous tactics, each specialised for specific proof generation methods. Among other things, they search mathematical libraries, simplify expressions, and identify lemmas and hypotheses to make progress in proofs. The proofs generated by this overlying programming machinery are terms in the core logical calculus which are checked mechanically by the proof checker. But in a large project, there is more to trust. It helps to understand the nature and limits of trust one can place on ITPs.
```
Osalotioman:
```suggestion
At its core, an interactive theorem prover implements an expressive logic, encoded in a suitable choice of mathematical foundations. \emph{Lean} 4 has the calculus of constructions extended by inductive types as its core logic. This logic is sufficiently powerful to express mathematical definitions and theorems for almost all areas of mathematical interest, while being relatively small and easy to write proof checkers for. Additionally, modern ITPs provide a convenient programming language which helps express mathematical ideas in a syntax closer to a mathematician's intuition than would be permitted by raw logical terms. A subset of ITPs like \emph{Lean}, \emph{Rocq}, and \emph{Isabelle} go one step further and provide the means to generate proofs through so-called \emph{tactics}. There are usually numerous tactics, each specialised for specific proof generation methods. Among other things, they search mathematical libraries, simplify expressions, and identify lemmas and hypotheses to make progress in proofs. The proofs generated by this overlying programming machinery are terms in the core logical calculus which are checked mechanically by the proof checker. But in a large project, there is more to trust. It helps to understand the nature and limits of trust one can place on ITPs.
```

Osalotioman:
`Of these three pieces, \emph{the kernel is the smallest and most trusted piece}. Programs in the programming language are translated by an \emph{elaborator} to the language of the kernel. In reality the elaborator does much more, but the key takeaways are the following:`

Earlier in the statement we already emphasized that the kernel is small, so I don't think there is a need to add `small` again when referring to the kernel here.

Shreyas4991:
I just noted one mistake I made. Isabelle doesn't generate proof terms unlike Lean and Rocq. 

Isabelle has a small kernel like Lean and Rocq, but relies on running the derivation rules of its logical theory (at least ideally speaking).

